PR #10, opened 2026-05-18
spec: Linux system packages meta architecture

### LINKED ISSUE #5: Port packaging over the open-telemetry/opentelemetry-injector
Revamp https://github.com/open-telemetry/opentelemetry-injector/pull/239 and bring over to this repository depending the findings of the build infrastructure from #4.

mmanciop:
When looking into my notes about the packages architecture (which packages we want to ship, which meta packages, which virtual, dependencies, etc.), I found out I had cut corners in the [PoC](https://github.com/open-telemetry/opentelemetry-injector/pull/239) in terms of implementing the virtual packages for languages, and that I had missed out entirely on future-proofing against "interface" changes among packages. The design document in this PR is intended to fill this gap.

Closes #5

### DIFF
--- a/docs/design/packages-meta-architecture.md
+++ b/docs/design/packages-meta-architecture.md
@@ -76,7 +76,7 @@ See [Injector interface versioning](#injector-interface-versioning) for the full
 **Swappable alternatives with interface versioning (`opentelemetry-java-autoinstrumentation1`, etc.).**
 Tracks generation 1 of the contract between the injector and each language's auto-instrumentation provider — the conf.d key names, the file layout under `/usr/lib/opentelemetry/<language>/`, and the `otel-config.yaml` structure.
 The metapackage depends on these virtual names.
-A vendor can ship a replacement package that also provides the same virtual name — combined with `Conflicts`/`Replaces` on the concrete upstream name, the package manager handles the swap transparently.
+A vendor can ship a replacement package with a different name (e.g., `acme-java-autoinstrumentation`) that also provides the same virtual name — combined with `Conflicts`/`Replaces` on the concrete upstream name, the package manager handles the swap transparently.
 If the upstream changes what "being a Java auto-instrumentation provider" means, it bumps to `opentelemetry-java-autoinstrumentation2`; existing vendor packages that still provide `…1` cannot satisfy the new dependency.
 See [Vendor Override](#vendor-override) for the recipe and user experience.
 
@@ -315,6 +315,17 @@ Each upstream language package declares a virtual package via `--provides` that 
 | `opentelemetry-java-autoinstrumentation` | `opentelemetry-java-autoinstrumentation1` |
 | `opentelemetry-nodejs-autoinstrumentation` | `opentelemetry-nodejs-autoinstrumentation1` |
 | `opentelemetry-dotnet-autoinstrumentation` | `opentelemetry-dotnet-autoinstrumentation1` |
+
+### Vendor package naming
+
+A vendor package **must** use a different concrete package name than the upstream package it replaces — for example, `acme-java-autoinstrumentation` rather than `opentelemetry-java-autoinstrumentation`.
+The vendor package then declares `Provides: opentelemetry-java-autoinstrumentation1` so that it satisfies the same virtual dependency, and `Conflicts`/`Replaces` on the upstream concrete name so the package manager handles the swap.
+
+| Upstream package | Vendor package (example) | Virtual package |
+|---|---|---|
+| `opentelemetry-java-autoinstrumentation` | `acme-java-autoinstrumentation` | `opentelemetry-java-autoinstrumentation1` |
+| `opentelemetry-nodejs-autoinstrumentation` | `acme-nodejs-autoinstrumentation` | `opentelemetry-nodejs-autoinstrumentation1` |
+| `opentelemetry-dotnet-autoinstrumentation` | `acme-dotnet-autoinstrumentation` | `opentelemetry-dotnet-autoinstrumentation1` |
 
 ### Vendor package recipe
 
--- a/docs/design/packages-meta-architecture.md
+++ b/docs/design/packages-meta-architecture.md
@@ -1,0 +1,602 @@
+# Linux System Packages Meta Architecture
+
+## Status
+
+Proposed
+
+## Context
+
+The Packaging SIG aims to deliver a streamlined, product-like experience for monitoring applications on virtual Linux hosts.
+Users should be able to achieve full auto-instrumentation with a single command:
+
+```sh
+{apt|yum} install opentelemetry
+```
+
+The SIG's [design principles](https://github.com/open-telemetry/community/blob/main/projects/packaging.md) further require:
+
+> Enable vendor packages to provide alternatives to upstream offerings
+
+A vendor (e.g., a commercial observability provider or a Linux distribution) should be able to ship their own Java, Node.js, or .NET auto-instrumentation package that cleanly replaces the upstream equivalent.
+Users who installed everything through the `opentelemetry` metapackage should be able to swap a single language package without breaking their system.
+
+This document describes all packages in the first version of the system packages, with the changes needed to support vendor overrides.
+
+> [!NOTE]
+> The current [POC](https://github.com/open-telemetry/opentelemetry-injector/pull/239) in the OpenTelemetry Injector repository does not yet support vendor overrides or interface versions. The [Current POC Gaps](#current-poc-gaps) section details the required changes.
+
+## Packages Overview
+
+The first version ships five packages.
+All are built with [FPM](https://fpm.readthedocs.io/) for both DEB and RPM.
+
+| Package | Description | Architecture |
+|---------|-------------|-------------|
+| `opentelemetry-injector` | LD_PRELOAD-based shared library that activates language agents | Per-arch (`amd64`, `arm64`) |
+| `opentelemetry-java-autoinstrumentation` | OpenTelemetry Java agent JAR | `all` / `noarch` |
+| `opentelemetry-nodejs-autoinstrumentation` | OpenTelemetry Node.js auto-instrumentation | `all` / `noarch` |
+| `opentelemetry-dotnet-autoinstrumentation` | OpenTelemetry .NET auto-instrumentation (glibc + musl) | Per-arch (`amd64`, `arm64`) |
+| `opentelemetry` | Metapackage that pulls in the injector and all language packages | `all` / `noarch` |
+
+### Dependency graph
+
+```
+opentelemetry  (metapackage)
+├── opentelemetry-injector1                      (virtual)
+├── opentelemetry-java-autoinstrumentation1      (virtual)
+├── opentelemetry-nodejs-autoinstrumentation1    (virtual)
+└── opentelemetry-dotnet-autoinstrumentation1    (virtual)
+
+opentelemetry-injector
+└── Provides: opentelemetry-injector1
+
+opentelemetry-java-autoinstrumentation
+├── Provides: opentelemetry-java-autoinstrumentation1
+└── Suggests: opentelemetry-injector1
+
+opentelemetry-nodejs-autoinstrumentation
+├── Provides: opentelemetry-nodejs-autoinstrumentation1
+└── Suggests: opentelemetry-injector1
+
+opentelemetry-dotnet-autoinstrumentation
+├── Provides: opentelemetry-dotnet-autoinstrumentation1
+└── Suggests: opentelemetry-injector1
+```
+
+Every dependency in the graph uses a virtual package name rather than a concrete one.
+The trailing `1` is not a package release version — it is the **interface generation number**, following the shared-library SONAME convention (`libssl3`, `libgcc-s1`).
+Both patterns are well-established in DEB and RPM (see [Appendix: Prior Art in DEB and RPM](#appendix-prior-art-in-deb-and-rpm)).
+
+**Interface versioning (`opentelemetry-injector1`).**
+Tracks generation 1 of the injector's `conf.d/` configuration API.
+The metapackage depends on this virtual name to ensure the installed injector is compatible with the language packages it pulls in.
+If a future injector release breaks the conf.d contract, it bumps to `opentelemetry-injector2`; the package manager blocks incompatible combinations automatically.
+See [Injector interface versioning](#injector-interface-versioning) for the full upgrade scenario.
+
+**Swappable alternatives with interface versioning (`opentelemetry-java-autoinstrumentation1`, etc.).**
+Tracks generation 1 of the contract between the injector and each language's auto-instrumentation provider — the conf.d key names, the file layout under `/usr/lib/opentelemetry/<language>/`, and the `otel-config.yaml` structure.
+The metapackage depends on these virtual names.
+A vendor can ship a replacement package that also provides the same virtual name — combined with `Conflicts`/`Replaces` on the concrete upstream name, the package manager handles the swap transparently.
+If the upstream changes what "being a Java auto-instrumentation provider" means, it bumps to `opentelemetry-java-autoinstrumentation2`; existing vendor packages that still provide `…1` cannot satisfy the new dependency.
+See [Vendor Override](#vendor-override) for the recipe and user experience.
+
+## Filesystem Layout
+
+All paths follow the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html) (FHS).
+
+```
+/usr/lib/opentelemetry/
+├── injector/
+│   └── libotelinject.so
+├── java/
+│   └── opentelemetry-javaagent.jar
+├── nodejs/
+│   └── node_modules/@opentelemetry/auto-instrumentations-node/…
+└── dotnet/
+    ├── glibc/…
+    └── musl/…
+
+/etc/opentelemetry/
+├── injector/
+│   ├── otelinject.conf
+│   ├── default_env.conf
+│   └── conf.d/
+│       ├── java.conf
+│       ├── nodejs.conf
+│       └── dotnet.conf
+├── java/
+│   └── otel-config.yaml
+├── nodejs/
+│   └── otel-config.yaml
+└── dotnet/
+    └── otel-config.yaml
+
+/usr/share/man/
+├── man8/opentelemetry-injector.8.gz
+└── man1/
+    ├── opentelemetry-java.1.gz
+    ├── opentelemetry-nodejs.1.gz
+    └── opentelemetry-dotnet.1.gz
+
+/usr/share/doc/
+├── opentelemetry-injector/
+├── opentelemetry-java-autoinstrumentation/
+├── opentelemetry-nodejs-autoinstrumentation/
+├── opentelemetry-dotnet-autoinstrumentation/
+└── opentelemetry/
+```
+
+## Package Definitions
+
+### `opentelemetry-injector`
+
+The core package.
+Installs `libotelinject.so`, a shared library loaded into every process via `/etc/ld.so.preload`.
+At runtime, the library inspects each process to determine if it is a Java, Node.js, or .NET application and, if so, activates the corresponding auto-instrumentation agent whose path is configured in the `conf.d/` drop-in files.
+
+#### Contents
+
+| Path | Description |
+|------|-------------|
+| `/usr/lib/opentelemetry/injector/libotelinject.so` | Injector shared library (per-arch) |
+| `/etc/opentelemetry/injector/otelinject.conf` | Main configuration file |
+| `/etc/opentelemetry/injector/default_env.conf` | Default `OTEL_*` environment variables for all agents |
+| `/etc/opentelemetry/injector/conf.d/` | Drop-in directory for language agent paths (empty until language packages are installed) |
+| `/usr/share/man/man8/opentelemetry-injector.8.gz` | Man page |
+| `/usr/share/doc/opentelemetry-injector/` | Documentation and copyright |
+
+#### Package metadata
+
+| Field | DEB | RPM |
+|-------|-----|-----|
+| Architecture | `amd64` or `arm64` | `x86_64` or `aarch64` |
+| Provides | `opentelemetry-injector1` | `opentelemetry-injector1` |
+| Config files | `/etc/opentelemetry/injector` | `/etc/opentelemetry/injector` |
+| Post-install | Appends `libotelinject.so` to `/etc/ld.so.preload` | Same |
+| Pre-uninstall | Removes `libotelinject.so` from `/etc/ld.so.preload` | Same |
+
+The post-install and pre-uninstall scripts must use only POSIX shell builtins (`read`, `case`, shell redirection) to avoid dependencies on `sed` or `grep`.
+
+#### Injector interface versioning
+
+The injector declares `Provides: opentelemetry-injector1`.
+The trailing `1` is not the package release version — it is the **generation number of the conf.d configuration API**, following the Debian shared-library naming convention (`libssl3`, `libgcc-s1`, `libpng16-16`).
+
+The metapackage depends on `opentelemetry-injector1` (hard dependency) to ensure the installed injector is compatible with the language packages it pulls in.
+Language packages only suggest `opentelemetry-injector1` (DEB `Suggests`; omitted on RPM which has no equivalent) since they are useful on their own without the injector.
+This decouples the API contract from the package's release cadence:
+
+- **Today:** the injector provides `opentelemetry-injector1`. The metapackage depends on it. Everything resolves.
+- **If a future injector release breaks the conf.d contract:** that release stops providing `opentelemetry-injector1` and starts providing `opentelemetry-injector2`. The metapackage still depends on `opentelemetry-injector1`, so the package manager blocks the upgrade until the metapackage (and the language packages it references) are also updated.
+- **Updated metapackage and language packages** switch to `opentelemetry-injector2`, and the system can upgrade atomically.
+
+This mechanism is self-service for vendors: a vendor package suggests `opentelemetry-injector1` and is automatically protected from incompatible injector upgrades without the injector needing to know the vendor package exists.
+
+### `opentelemetry-java-autoinstrumentation`
+
+Downloads the upstream [OpenTelemetry Java agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) JAR at build time and packages it.
+
+#### Contents
+
+| Path | Description |
+|------|-------------|
+| `/usr/lib/opentelemetry/java/opentelemetry-javaagent.jar` | Java agent JAR |
+| `/etc/opentelemetry/injector/conf.d/java.conf` | Drop-in: `jvm_auto_instrumentation_agent_path=/usr/lib/opentelemetry/java/opentelemetry-javaagent.jar` |
+| `/etc/opentelemetry/java/otel-config.yaml` | Declarative configuration template |
+| `/usr/share/man/man1/opentelemetry-java.1.gz` | Man page |
+| `/usr/share/doc/opentelemetry-java-autoinstrumentation/` | Documentation and copyright |
+
+#### Package metadata
+
+| Field | DEB | RPM |
+|-------|-----|-----|
+| Architecture | `all` | `noarch` |
+| Provides | `opentelemetry-java-autoinstrumentation1` | `opentelemetry-java-autoinstrumentation1` |
+| Suggests | `opentelemetry-injector1`, `default-jre` (DEB only) | — |
+| Config files | `/etc/opentelemetry/java` | `/etc/opentelemetry/java` |
+
+### `opentelemetry-nodejs-autoinstrumentation`
+
+Downloads [`@opentelemetry/auto-instrumentations-node`](https://www.npmjs.com/package/@opentelemetry/auto-instrumentations-node) from npm at build time and packages the installed `node_modules` tree.
+
+#### Contents
+
+| Path | Description |
+|------|-------------|
+| `/usr/lib/opentelemetry/nodejs/node_modules/…` | Node.js auto-instrumentation modules |
+| `/etc/opentelemetry/injector/conf.d/nodejs.conf` | Drop-in: `nodejs_auto_instrumentation_agent_path=…/register.js` |
+| `/etc/opentelemetry/nodejs/otel-config.yaml` | Declarative configuration template |
+| `/usr/share/man/man1/opentelemetry-nodejs.1.gz` | Man page |
+| `/usr/share/doc/opentelemetry-nodejs-autoinstrumentation/` | Documentation and copyright |
+
+#### Package metadata
+
+| Field | DEB | RPM |
+|-------|-----|-----|
+| Architecture | `all` | `noarch` |
+| Provides | `opentelemetry-nodejs-autoinstrumentation1` | `opentelemetry-nodejs-autoinstrumentation1` |
+| Suggests | `opentelemetry-injector1`, `nodejs (>= 18)` (DEB only) | — |
+| Config files | `/etc/opentelemetry/nodejs` | `/etc/opentelemetry/nodejs` |
+
+### `opentelemetry-dotnet-autoinstrumentation`
+
+Downloads the [OpenTelemetry .NET auto-instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation) binaries at build time, for both glibc and musl libc flavors.
+The injector detects the libc flavor at runtime by reading ELF headers.
+
+#### Contents
+
+| Path | Description |
+|------|-------------|
+| `/usr/lib/opentelemetry/dotnet/glibc/…` | .NET agent binaries (glibc) |
+| `/usr/lib/opentelemetry/dotnet/musl/…` | .NET agent binaries (musl) |
+| `/etc/opentelemetry/injector/conf.d/dotnet.conf` | Drop-in: `dotnet_auto_instrumentation_agent_path_prefix=/usr/lib/opentelemetry/dotnet` |
+| `/etc/opentelemetry/dotnet/otel-config.yaml` | Declarative configuration template |
+| `/usr/share/man/man1/opentelemetry-dotnet.1.gz` | Man page |
+| `/usr/share/doc/opentelemetry-dotnet-autoinstrumentation/` | Documentation and copyright |
+
+#### Package metadata
+
+| Field | DEB | RPM |
+|-------|-----|-----|
+| Architecture | `amd64` or `arm64` | `x86_64` or `aarch64` |
+| Provides | `opentelemetry-dotnet-autoinstrumentation1` | `opentelemetry-dotnet-autoinstrumentation1` |
+| Suggests | `opentelemetry-injector1` (DEB only) | — |
+| Config files | `/etc/opentelemetry/dotnet` | `/etc/opentelemetry/dotnet` |
+
+### `opentelemetry`
+
+A metapackage with no files of its own (besides a README under `/usr/share/doc/`).
+It exists so that `apt install opentelemetry` or `yum install opentelemetry` pulls in the full auto-instrumentation suite.
+
+#### Package metadata
+
+| Field | DEB | RPM |
+|-------|-----|-----|
+| Architecture | `all` | `noarch` |
+| Depends | `opentelemetry-injector1` | `opentelemetry-injector1` |
+| Depends | `opentelemetry-java-autoinstrumentation1` | `opentelemetry-java-autoinstrumentation1` |
+| Depends | `opentelemetry-nodejs-autoinstrumentation1` | `opentelemetry-nodejs-autoinstrumentation1` |
+| Depends | `opentelemetry-dotnet-autoinstrumentation1` | `opentelemetry-dotnet-autoinstrumentation1` |
+
+Every dependency uses a virtual name rather than a concrete package name.
+`opentelemetry-injector1` ensures compatibility with the injector's conf.d API generation.
+Language package dependencies (`opentelemetry-<language>-autoinstrumentation1`) serve a dual purpose: they version the provider interface *and* allow either the upstream or a vendor package to satisfy them.
+See [Injector interface versioning](#injector-interface-versioning) for the upgrade scenario.
+
+## Configuration System
+
+### Hierarchy
+
+1. **`/etc/opentelemetry/injector/otelinject.conf`** — main injector configuration. Points to the default environment file and documents the `conf.d/` mechanism.
+
+2. **`/etc/opentelemetry/injector/default_env.conf`** — default `OTEL_*` environment variables applied to all instrumented processes. Format: `KEY=VALUE`, one per line.
+
+3. **`/etc/opentelemetry/injector/conf.d/*.conf`** — drop-in files read in alphabetical order. Each language package installs one file that sets the agent path. Later files override earlier ones for the same key.
+
+4. **`/etc/opentelemetry/<language>/otel-config.yaml`** — per-language declarative configuration templates (used when `OTEL_EXPERIMENTAL_CONFIG_FILE` is set).
+
+### Available conf.d settings
+
+| Key | Set by | Description |
+|-----|--------|-------------|
+| `jvm_auto_instrumentation_agent_path` | `conf.d/java.conf` | Absolute path to the Java agent JAR |
+| `nodejs_auto_instrumentation_agent_path` | `conf.d/nodejs.conf` | Absolute path to the Node.js agent entry point |
+| `dotnet_auto_instrumentation_agent_path_prefix` | `conf.d/dotnet.conf` | Directory prefix for the .NET agent (injector appends `glibc/` or `musl/`) |
+| `all_auto_instrumentation_agents_env_path` | `otelinject.conf` | Path to the default environment variables file |
+
+## File Ownership Boundaries
+
+Each package owns a disjoint set of paths.
+This is critical for `Conflicts`/`Replaces` to work correctly.
+
+| Package | Owns |
+|---------|------|
+| `opentelemetry-injector` | `/usr/lib/opentelemetry/injector/`, `/etc/opentelemetry/injector/otelinject.conf`, `/etc/opentelemetry/injector/default_env.conf`, `/etc/opentelemetry/injector/conf.d/` (directory only) |
+| `opentelemetry-java-autoinstrumentation` | `/usr/lib/opentelemetry/java/`, `/etc/opentelemetry/injector/conf.d/java.conf`, `/etc/opentelemetry/java/` |
+| `opentelemetry-nodejs-autoinstrumentation` | `/usr/lib/opentelemetry/nodejs/`, `/etc/opentelemetry/injector/conf.d/nodejs.conf`, `/etc/opentelemetry/nodejs/` |
+| `opentelemetry-dotnet-autoinstrumentation` | `/usr/lib/opentelemetry/dotnet/`, `/etc/opentelemetry/injector/conf.d/dotnet.conf`, `/etc/opentelemetry/dotnet/` |
+| `opentelemetry` | `/usr/share/doc/opentelemetry/` |
+
+A vendor replacement package *must* own the same set of paths as the upstream package it replaces.
+This is what makes `--replaces` work: `dpkg` allows the new package to overwrite files owned by the replaced package.
+
+## Vendor Override
+
+### Virtual package names
+
+Each upstream language package declares a virtual package via `--provides` that any alternative provider can also satisfy:
+
+| Concrete package | Virtual package |
+|---|---|
+| `opentelemetry-java-autoinstrumentation` | `opentelemetry-java-autoinstrumentation1` |
+| `opentelemetry-nodejs-autoinstrumentation` | `opentelemetry-nodejs-autoinstrumentation1` |
+| `opentelemetry-dotnet-autoinstrumentation` | `opentelemetry-dotnet-autoinstrumentation1` |
+
+### Vendor package recipe
+
+A vendor building a replacement for, say, the Java auto-instrumentation package would use:
+
+```bash
+# DEB
+fpm -s dir -t deb -n "acme-java-autoinstrumentation" -v "$ACME_VERSION" \
+    --provides "opentelemetry-java-autoinstrumentation1" \
+    --conflicts "opentelemetry-java-autoinstrumentation" \
+    --replaces "opentelemetry-java-autoinstrumentation" \
+    --deb-suggests "opentelemetry-injector1" \
+    --config-files /etc/opentelemetry/injector/conf.d/ \
+    --config-files /etc/opentelemetry/java/ \
+    …
+```
+
+```bash
+# RPM (no Suggests equivalent — the hint is omitted)
+fpm -s dir -t rpm -n "acme-java-autoinstrumentation" -v "$ACME_VERSION" \
+    --provides "opentelemetry-java-autoinstrumentation1" \
+    --conflicts "opentelemetry-java-autoinstrumentation" \
+    --replaces "opentelemetry-java-autoinstrumentation" \
+    --config-files /etc/opentelemetry/injector/conf.d/ \
+    --config-files /etc/opentelemetry/java/ \
+    …
+```
+
+Key properties:
+
+- `--provides` the virtual name so the metapackage's dependency is satisfied.
+- `--conflicts` and `--replaces` the concrete upstream name so the package manager removes the upstream package automatically during install.
+- The vendor package installs its own `conf.d/java.conf` (same filename, not a higher-priority override) pointing to its agent path. Because it `--replaces` the upstream package, the upstream's `conf.d/java.conf` is cleanly removed by the package manager.
+- The injector is a `Suggests` (DEB only), not a hard dependency. Language packages are useful on their own (e.g., a Java agent JAR can be used directly via `-javaagent:`), and the metapackage already ensures co-installation for the standard use case. RPM has no equivalent of `Suggests`, so the hint is omitted there.
+
+### Conf.d file naming convention
+
+Vendor packages should use the **same conf.d filename** as the upstream package they replace (e.g., `java.conf`, not `99-vendor-java.conf`).
+This ensures:
+
+1. Only one config file per language is present at any time.
+2. No ordering ambiguity.
+3. The package manager handles the file transition via `Replaces`.
+
+Custom user overrides (not managed by any package) should use a numeric prefix to sort after package-managed files: e.g., `99-local-java.conf`.
+
+### User experience
+
+#### Fresh install with upstream packages only
+
+```sh
+apt install opentelemetry
+```
+
+Installs the metapackage and all upstream language packages.
+Every `opentelemetry-<language>-autoinstrumentation1` dependency is satisfied by the corresponding concrete upstream package.
+
+#### Fresh install with a vendor package
+
+```sh
+apt install opentelemetry acme-java-autoinstrumentation
+```
+
+`apt` resolves `opentelemetry-java-autoinstrumentation1` via `acme-java-autoinstrumentation`'s `Provides`.
+Because `acme-java-autoinstrumentation` declares `Conflicts: opentelemetry-java-autoinstrumentation`, `apt` skips the upstream Java package entirely.
+The remaining language packages (Node.js, .NET) are installed from upstream as usual.
+
+#### Swapping on an existing system
+
+```sh
+apt install acme-java-autoinstrumentation
+```
+
+`apt` sees `Conflicts: opentelemetry-java-autoinstrumentation` and `Replaces: opentelemetry-java-autoinstrumentation`, removes the upstream package, installs the vendor package.
+The metapackage remains installed because its dependency on `opentelemetry-java-autoinstrumentation1` is still satisfied.
+
+#### Reverting to upstream
+
+```sh
+apt install opentelemetry-java-autoinstrumentation
+```
+
+`dpkg` removes the vendor package (symmetric `Conflicts`/`Replaces` if the vendor chose to declare them, or the user runs `apt remove acme-java-autoinstrumentation` first) and installs upstream.
+
+## Current POC Gaps
+
+The [POC](https://github.com/open-telemetry/opentelemetry-injector/pull/239) in the OpenTelemetry Injector repository implements most of the architecture above but has the following gaps relative to the proposed design:
+
+### 1. No interface version on the injector
+
+The injector package does not declare `Provides: opentelemetry-injector1`.
+There is no mechanism to prevent an incompatible injector upgrade from breaking installed language packages.
+All consumers depend on the concrete package name (`opentelemetry-injector`) instead of a versioned interface name.
+
+### 2. No interface version on language packages
+
+Language packages do not declare `Provides` with a versioned virtual name (e.g., `opentelemetry-java-autoinstrumentation1`).
+There is no abstraction that both the upstream and a vendor package can satisfy, and no way to detect a breaking change in the provider contract.
+
+### 3. Metapackage depends on concrete package names with exact versions
+
+```bash
+# DEB
+--depends "opentelemetry-java-autoinstrumentation (= ${VERSION})"
+# RPM
+--depends "opentelemetry-java-autoinstrumentation = ${VERSION}"
+```
+
+A vendor package cannot satisfy these dependencies.
+Installing a vendor alternative forces the user to first remove the metapackage, which also removes any `apt autoremove` / `dnf autoremove` protection for the other language packages.
+
+### 4. No `Conflicts` / `Replaces` metadata
+
+If a vendor package installs files to the same paths under `/usr/lib/opentelemetry/<language>/` or `/etc/opentelemetry/injector/conf.d/`, `dpkg` and `rpm` will refuse the installation due to file conflicts.
+Neither the upstream nor the vendor package declares ownership boundaries.
+
+### 5. Unnecessary `sed` and `grep` dependencies
+
+The injector's post-install and pre-uninstall scripts use `grep` and `sed` to manipulate `/etc/ld.so.preload`, causing the package to declare `--depends sed` and `--depends grep`.
+While these tools are practically always present, the dependencies are unnecessary and should be eliminated by rewriting the scripts to use only POSIX shell builtins.
+
+### 6. Conf.d files cannot coexist
+
+If the upstream package owns `conf.d/java.conf` and the vendor package installs `conf.d/99-vendor-java.conf`, *both* config files are present.
+Because the injector reads files in alphabetical order and the last value wins, the vendor file's value takes precedence at runtime.
+But the upstream file still sets the path on every read before the vendor file overrides it, creating a fragile ordering dependency.
+Worse, the upstream JAR is still on disk consuming space, and the upstream package is still installed, making `dpkg -l` / `rpm -qa` output misleading.
+
+## Required Changes to the POC
+
+### Injector build scripts
+
+| File | Change |
+|------|--------|
+| `packaging/deb/injector/build.sh` | Add `--provides "opentelemetry-injector1"`, remove `--depends sed` and `--depends grep` |
+| `packaging/rpm/injector/build.sh` | Add `--provides "opentelemetry-injector1"`, remove `--depends sed` and `--depends grep` |
+| `packaging/common/scripts/postinstall-injector.sh` | Rewrite to use POSIX shell builtins only (no `grep`) |
+| `packaging/common/scripts/preuninstall-injector.sh` | Rewrite to use POSIX shell builtins only (no `grep`, `sed`) |
+
+### Language package build scripts (DEB)
+
+| File | Change |
+|------|--------|
+| `packaging/deb/java/build.sh` | Add `--provides "opentelemetry-java-autoinstrumentation1"`, replace `--depends "opentelemetry-injector (>= ${VERSION})"` with `--deb-suggests "opentelemetry-injector1"` |
+| `packaging/deb/nodejs/build.sh` | Add `--provides "opentelemetry-nodejs-autoinstrumentation1"`, replace `--depends "opentelemetry-injector (>= ${VERSION})"` with `--deb-suggests "opentelemetry-injector1"` |
+| `packaging/deb/dotnet/build.sh` | Add `--provides "opentelemetry-dotnet-autoinstrumentation1"`, replace `--depends "opentelemetry-injector (>= ${VERSION})"` with `--deb-suggests "opentelemetry-injector1"` |
+
+### Language package build scripts (RPM)
+
+| File | Change |
+|------|--------|
+| `packaging/rpm/java/build.sh` | Add `--provides "opentelemetry-java-autoinstrumentation1"`, remove `--depends "opentelemetry-injector >= ${VERSION}"` |
+| `packaging/rpm/nodejs/build.sh` | Add `--provides "opentelemetry-nodejs-autoinstrumentation1"`, remove `--depends "opentelemetry-injector >= ${VERSION}"` |
+| `packaging/rpm/dotnet/build.sh` | Add `--provides "opentelemetry-dotnet-autoinstrumentation1"`, remove `--depends "opentelemetry-injector >= ${VERSION}"` |
+
+RPM has no equivalent of `Suggests`, so the injector hint is omitted there.
+
+### Metapackage build scripts
+
+| File | Change |
+|------|--------|
+| `packaging/deb/meta/build.sh` | Change `--depends "opentelemetry-injector (= ${VERSION})"` to `--depends "opentelemetry-injector1"`, change `--depends "opentelemetry-java-autoinstrumentation (= ${VERSION})"` to `--depends "opentelemetry-java-autoinstrumentation1"` (same for nodejs, dotnet) |
+| `packaging/rpm/meta/build.sh` | Change `--depends "opentelemetry-injector = ${VERSION}"` to `--depends "opentelemetry-injector1"`, change `--depends "opentelemetry-java-autoinstrumentation = ${VERSION}"` to `--depends "opentelemetry-java-autoinstrumentation1"` (same for nodejs, dotnet) |
+
+### No changes needed
+
+- **Runtime configuration system**: the `conf.d/` mechanism already supports the override model. No code changes needed in the injector binary.
+- **Installation scripts** (`postinstall-injector.sh`, `preuninstall-injector.sh`): unaffected.
+
+## Alternatives Considered
+
+### Conf.d-only override (no package metadata changes)
+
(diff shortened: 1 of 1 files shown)


mmanciop:
Out of curiosity for whether AI would find the interface versioning idiomatic, I asked Claude:

> What is the trailing 1 in this Linux system deb package name? opentelemetry-injector1

Claude: In both .deb and RPM conventions, the trailing 1 is an ABI/major-version slot embedded directly in the package name. It signals “this is version-slot 1 of this component,” and its primary purpose is to allow parallel installation of incompatible major versions on the same system.

You see this pattern everywhere:

	•	libssl3 / libssl1.1 — two OpenSSL ABIs coexisting
	•	libpq5 — PostgreSQL client library, slot 5
	•	python3 / python2

For something like opentelemetry-injector1, the implication is:

	•	The 1 is not the package version (that lives in the version field, e.g. 1.2.3-1). It’s part of the package name itself.
	•	If a future release breaks ABI or makes incompatible changes, it ships as opentelemetry-injector2, and both can coexist on the same host without conflicts.
	•	Tooling, init scripts, or other packages that depend on a specific ABI pin to opentelemetry-injector1 explicitly.

In .deb land this is especially common for shared libraries (tied to the ELF SONAME), but it’s also used for non-library packages that inject into processes or have ABI-sensitive interfaces — which an OTel injector almost certainly does, since it likely ships a .so that gets LD_PRELOAD-ed or injected via /etc/ld.so.preload.

So the 1 is essentially a forward-compatibility escape hatch baked into the name.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+│   └── conf.d/
+│       ├── java.conf
+│       ├── nodejs.conf
+│       └── dotnet.conf
+├── java/
+│   └── otel-config.yaml
+├── nodejs/
+│   └── otel-config.yaml
+└── dotnet/
+    └── otel-config.yaml
```
thompson-tomo, line 125: Wouldn't it be better to use a single config file?

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+| `opentelemetry-injector` | `/usr/lib/opentelemetry/injector/`, `/etc/opentelemetry/injector/otelinject.conf`, `/etc/opentelemetry/injector/default_env.conf`, `/etc/opentelemetry/injector/conf.d/` (directory only) |
+| `opentelemetry-java-autoinstrumentation` | `/usr/lib/opentelemetry/java/`, `/etc/opentelemetry/injector/conf.d/java.conf`, `/etc/opentelemetry/java/` |
+| `opentelemetry-nodejs-autoinstrumentation` | `/usr/lib/opentelemetry/nodejs/`, `/etc/opentelemetry/injector/conf.d/nodejs.conf`, `/etc/opentelemetry/nodejs/` |
+| `opentelemetry-dotnet-autoinstrumentation` | `/usr/lib/opentelemetry/dotnet/`, `/etc/opentelemetry/injector/conf.d/dotnet.conf`, `/etc/opentelemetry/dotnet/` |
+| `opentelemetry` | `/usr/share/doc/opentelemetry/` |
+
+A vendor replacement package *must* own the same set of paths as the upstream package it replaces.
+This is what makes `--replaces` work: `dpkg` allows the new package to overwrite files owned by the replaced package.
+
+## Vendor Override
```
thompson-tomo: Thoughts about lowering the barrier to entry for vendors/companies by enabling declarative config to define the auto-instrumentation which is to be used.

The injector would use a path based on declarative if it is defined otherwise uses the conf file.

This would reduce the need for companies to be hosting their own packages.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+- **Updated metapackage and language packages** switch to `opentelemetry-injector2`, and the system can upgrade atomically.
+
+The same logic applies to language package interface generations. When the metapackage moves from `opentelemetry-java-autoinstrumentation1` to `opentelemetry-java-autoinstrumentation2`, the package manager upgrades the upstream language package in the same transaction.
+
+**Impact on vendor packages.** If a user has a vendor package that provides `opentelemetry-java-autoinstrumentation1` but the new metapackage requires `opentelemetry-java-autoinstrumentation2`, the package manager holds back the metapackage upgrade until the vendor ships an updated package providing `…2`. This is the intended safety behavior — it prevents a vendor package from being silently used with an incompatible interface — but it means the user cannot upgrade to the new metapackage until their vendor catches up.
+
+This mechanism is self-service for vendors: a vendor package provides a given interface generation and is automatically protected from incompatible upgrades without the upstream needing to know the vendor package exists.
+
+### `opentelemetry-java-autoinstrumentation`
```
atoulme, line 200: @open-telemetry/java-maintainers please review. You will be codeowners of this package configuration.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+#### Package metadata
+
+| Field | DEB | RPM |
+|-------|-----|-----|
+| Architecture | `all` | `noarch` |
+| Provides | `opentelemetry-java-autoinstrumentation1` | `opentelemetry-java-autoinstrumentation1` |
+| Suggests | `opentelemetry-injector1`, `default-jre` (DEB only) | — |
+| Config files | `/etc/opentelemetry/java` | `/etc/opentelemetry/java` |
+
+### `opentelemetry-nodejs-autoinstrumentation`
```
atoulme, line 224: @open-telemetry/javascript-maintainers please review. You will be codeowners of this package configuration.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+#### Package metadata
+
+| Field | DEB | RPM |
+|-------|-----|-----|
+| Architecture | `all` | `noarch` |
+| Provides | `opentelemetry-nodejs-autoinstrumentation1` | `opentelemetry-nodejs-autoinstrumentation1` |
+| Suggests | `opentelemetry-injector1`, `nodejs (>= 18)` (DEB only) | — |
+| Config files | `/etc/opentelemetry/nodejs` | `/etc/opentelemetry/nodejs` |
+
+### `opentelemetry-dotnet-autoinstrumentation`
```
atoulme, line 248: @open-telemetry/dotnet-maintainers please review. You will be codeowners of this package configuration.

martincostello: /cc @open-telemetry/dotnet-instrumentation-maintainers

docs/design/packages-meta-architecture.md
jack-berg: cc @open-telemetry/java-instrumentation-maintainers

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+
+Downloads the upstream [OpenTelemetry Java agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) JAR at build time and packages it.
+
+#### Contents
+
+| Path | Description |
+|------|-------------|
+| `/usr/lib/opentelemetry/java/opentelemetry-javaagent.jar` | Java agent JAR |
+| `/etc/opentelemetry/injector/conf.d/java.conf` | Drop-in: `jvm_auto_instrumentation_agent_path=/usr/lib/opentelemetry/java/opentelemetry-javaagent.jar` |
+| `/etc/opentelemetry/java/otel-config.yaml` | Declarative configuration template |
```
jack-berg: What's the convention for file naming for the collector? I want to avoid conflicts / confusion.

```suggestion
| `/etc/opentelemetry/java/otel-sdk-config.yaml` | Declarative configuration template |
```

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+
+### `opentelemetry-java-autoinstrumentation`
+
+Downloads the upstream [OpenTelemetry Java agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) JAR at build time and packages it.
+
+#### Contents
+
+| Path | Description |
+|------|-------------|
+| `/usr/lib/opentelemetry/java/opentelemetry-javaagent.jar` | Java agent JAR |
```
jack-berg, line 209: Naive question: How do I know which version of javaagent is used in this package at a glance?

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+
+**Impact on vendor packages.** If a user has a vendor package that provides `opentelemetry-java-autoinstrumentation1` but the new metapackage requires `opentelemetry-java-autoinstrumentation2`, the package manager holds back the metapackage upgrade until the vendor ships an updated package providing `…2`. This is the intended safety behavior — it prevents a vendor package from being silently used with an incompatible interface — but it means the user cannot upgrade to the new metapackage until their vendor catches up.
+
+This mechanism is self-service for vendors: a vendor package provides a given interface generation and is automatically protected from incompatible upgrades without the upstream needing to know the vendor package exists.
+
+### `opentelemetry-java-autoinstrumentation`
+
+Downloads the upstream [OpenTelemetry Java agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) JAR at build time and packages it.
```
jack-berg: Just to be clear, the package does not provide any utility for downloading the java agent, right?

What's the versioning scheme for the package? Hopefully its in lockstep with the otel java agent versions. That should probably be true in general, not just for java. Is that mentioned anywhere?

mmanciop: The package will contain the Jar file directly. Versioning for the packages is not something I thought about tbh. In case of Java, using the same version as the Jar file makes perfect sense. For pretty much all other language packages, they will carry a variety of artifacts inside with different versions: for the same language, SDK and contrib packages usually have different versions schemes.

docs/design/packages-meta-architecture.md
mmanciop: Not really, one of the most useful functionalities is going to be the enablement and disablement of instrumentations, and those are language-specific.

docs/design/packages-meta-architecture.md
mmanciop: I doubt vendors would like to ship a package containing only configurations. All distros I have seen (and build myself over the years), are code plus configurations. But I would be very interested in being proven wrong.

docs/design/packages-meta-architecture.md
mmanciop: By looking at the file system? We could add a sort of BOM file for that. I think we need to solve this together with the versioning question.

docs/design/packages-meta-architecture.md
mmanciop: The collector is not yet part of this system of packages, I remembered we had agreed to bring that in later.

docs/design/packages-meta-architecture.md
thompson-tomo: Vendors wouldn't ship a package containing only configurations and in fact they wouldn't ship an OS package at all. They would only ship their instrumentation package ie npm. The process to install it would be install the official otel package, add the custom instrumentation to your config which is then deployed using opamp to the instances, the opamp agent then installs it on the instances hence lowers the barrier to entry.

docs/design/packages-meta-architecture.md
thompson-tomo: Isn't that already handled given instrumentation config is already split based on language see https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md

Further how would the multiple config scenario work with opamp to deploy the config?

docs/design/packages-meta-architecture.md
mmanciop: In that case, the OpAMP would simply override the system package's then?

docs/design/packages-meta-architecture.md
mmanciop: > Isn't that already handled given instrumentation config is already split based on language see https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md

To my understanding, no, it is not. IIRC @jack-berg also had a remark that it is best to keep the configs for different languages separately as the behavior of trying to disable an instrumentation that does not exist for the SDK is very chancy.

> Further how would the multiple config scenario work with opamp to deploy the config?

Assuming SDKs get OpAMP support (which I never heard to be a thing tbh), then they configure to use OpAMP and it works as normal. For the collector, it's a different story, and that already gets today a separate config and separate system package.

mmanciop: Well it turns out Java does support OpAMP, so I take that back. Let me think about it a bit.

mmanciop: A quick query with Claude says that OpAMP is not something that can be configured via env vars or declarative config in upstream Java agent. I assume it will eventually be configurable via declarative config, so the answer will likely be "you change the declarative config to tell the SDK to fetch its actual configs via OpAMP". But in truth, it seems to me too early for any concrete answer to be reliable. @jack-berg maybe you have something to add here?

thompson-tomo: That would be correct as my understanding of opamp based on https://github.com/open-telemetry/opamp-spec/blob/main/specification.md

> Open Agent Management Protocol (OpAMP) is a network protocol for remote management of large fleets of data collection Agents.

Is that opamp is how to manage the configuration of the agents which is distributed to the agents & even handles deploying updates to them.

docs/design/packages-meta-architecture.md
thompson-tomo: Correct the agent specified in the config would override the builtin agent.

docs/design/packages-meta-architecture.md
mmanciop: @jack-berg given the fact that we do prefix the config file with the language, do you really think we need to add `-sdk` to the name? That would not conflict with the operator config file.

docs/design/packages-meta-architecture.md
mmanciop: Created #13 to track this

docs/design/packages-meta-architecture.md
mmanciop: > Vendors wouldn't ship a package containing only configurations and in fact they wouldn't ship an OS package at all. They would only ship their instrumentation package ie npm. The process to install it would be install the official otel package, add the custom instrumentation to your config which is then deployed using opamp to the instances, the opamp agent then installs it on the instances hence lowers the barrier to entry.

I honestly do not think this is a good idea. If the vendor wants to make a distro and have it consumed by autoinstrumentations delivered and activated by system packages, they package the distro as a system package that replaces the community one. It could technically be possible for vendors to create independent system packages and have users install them and then modify the configs of the community packages to have those instrumentations additionally activated. These would look very similar to the node dependencies you find in Ubuntu's `main`. (And they are a terrible experience for package maintainers, I am told.) About activating those additional instrumentations by config, it is expected for users to modify the declarative configurations we ship (which would also be the way to activate OpAMP, by the way), as with pretty much any file living in `/etc`. But this all would be such a cumbersome process for adopters that I really don't think makes sense to consider.

docs/design/packages-meta-architecture.md
mmanciop: In my experience, it never ends well when updates via systems packages intermingle with updates through other channels.

We should not mix the update process of system packages and that of OpAMP. If end users want to distribute instrumentations through OpAMP, then they should do that instead of system packages, I think.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+
+> [!NOTE]
+> The current [POC](https://github.com/open-telemetry/opentelemetry-injector/pull/239) in the OpenTelemetry Injector repository does not yet support vendor overrides or interface versions. The [Current POC Gaps](#current-poc-gaps) section details the required changes.
+
+## Packages Overview
+
+The first version ships five packages.
+All are built with [FPM](https://fpm.readthedocs.io/) for both DEB and RPM.
+
+| Package | Description | Architecture |
```
Kielek, line 33: I think that you should use here official names. Putting value for .NET Auto only.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+|---------|-------------|-------------|
+| `opentelemetry-injector` | LD_PRELOAD-based shared library that activates language agents | Per-arch (`amd64`, `arm64`) |
+| `opentelemetry-java-autoinstrumentation` | OpenTelemetry Java agent JAR | `all` / `noarch` |
+| `opentelemetry-nodejs-autoinstrumentation` | OpenTelemetry Node.js auto-instrumentation | `all` / `noarch` |
+| `opentelemetry-dotnet-autoinstrumentation` | OpenTelemetry .NET auto-instrumentation (glibc + musl) | Per-arch (`amd64`, `arm64`) |
```
Kielek:
```suggestion
| `opentelemetry-dotnet-autoinstrumentation` | OpenTelemetry .NET Automatic Instrumentation (glibc + musl) | Per-arch (`amd64`, `arm64`) |
```

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+| Field | DEB | RPM |
+|-------|-----|-----|
+| Architecture | `all` | `noarch` |
+| Provides | `opentelemetry-nodejs-autoinstrumentation1` | `opentelemetry-nodejs-autoinstrumentation1` |
+| Suggests | `opentelemetry-injector1`, `nodejs (>= 18)` (DEB only) | — |
+| Config files | `/etc/opentelemetry/nodejs` | `/etc/opentelemetry/nodejs` |
+
+### `opentelemetry-dotnet-autoinstrumentation`
+
+Downloads the [OpenTelemetry .NET auto-instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation) binaries at build time, for both glibc and musl libc flavors.
```
Kielek: Naming

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+The injector detects the libc flavor at runtime by reading ELF headers.
+
+#### Contents
+
+| Path | Description |
+|------|-------------|
+| `/usr/lib/opentelemetry/dotnet/glibc/…` | .NET agent binaries (glibc) |
+| `/usr/lib/opentelemetry/dotnet/musl/…` | .NET agent binaries (musl) |
```
Kielek: I think that you will have a lot of duplication.
If I remember correctly, the differences are only in one library. Other files are common.
It will work, but you have potentially bigger size than needed.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+| `/etc/opentelemetry/injector/conf.d/dotnet.conf` | Drop-in: `dotnet_auto_instrumentation_agent_path_prefix=/usr/lib/opentelemetry/dotnet` |
+| `/etc/opentelemetry/dotnet/otel-config.yaml` | Declarative configuration template |
+| `/usr/share/man/man1/opentelemetry-dotnet.1.gz` | Man page |
+| `/usr/share/doc/opentelemetry-dotnet-autoinstrumentation/` | Documentation and copyright |
+
+#### Package metadata
+
+| Field | DEB | RPM |
+|-------|-----|-----|
+| Architecture | `amd64` or `arm64` | `x86_64` or `aarch64` |
```
Kielek, line 268: Keep in mind, that the latest version supports the same set of systems as the .NET team: https://dotnet.microsoft.com/en-us/download/dotnet

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+| Provides | `opentelemetry-dotnet-autoinstrumentation1` | `opentelemetry-dotnet-autoinstrumentation1` |
```
Kielek, line 269: `1` value, 1 version can be breaking change in terms of supported set of .NET versions.
Current version supports .NET 8. 9, 10, In the November we will go to .NET 10 and 11 only (8 and 9 will be no longer supported).
I am not sure if it has any impact on your plans.

docs/design/packages-meta-architecture.md
thompson-tomo: > it never ends well when updates via systems packages intermingle with updates through other channels.
> 
> We should not mix the update process of system packages and that of OpAMP. If end users want to distribute instrumentations through OpAMP, then they should do that instead of system packages, I think.

I am not suggesting to mix the update process of the system package and in fact I would keep it consistent across all use cases. The difference would be if someone wants to switch from the builtin to a custom they define it in declarative config.

mmanciop: That would work, in the declarative config of the injector. The `/etc/opentelemetry/injector/<language>.conf` specifies where the SDK and autoinstrumentations are found, but they need to be a full setup (SDK + autoinstrumentations); how that looks like changes from language to language.

docs/design/packages-meta-architecture.md
thompson-tomo: Yes vendors are more likely to make a system package  as they have a range of customers who would use it hence a public repository/registry works but that is not the only use-case. The other use case would be Companies who build their own auto instrumentation distribution with additional instrumentations.

> It could technically be possible for vendors to create independent system packages and have users install them and then modify the configs of the community packages to have those instrumentations additionally activated.

I didn't suggest that and wouldn't recommend it, the suggestion was define in declarative config an override of the auto instrumentation which is to be used.

> About activating those additional instrumentations by config, it is expected for users to modify the declarative configurations we ship (which would also be the way to activate OpAMP, by the way), as with pretty much any file living in /etc. But this all would be such a cumbersome process for adopters that I really don't think makes sense to consider.

I think I see where the confusion is, I am invisioning using opamp to configure the system not just the language agent. This enables a non cumbersome solution which is scalable from a few instances to many. The process for users would be the following:

- install opentelemetry system package
- configure opamp address
- the configuration is downloaded from the opamp server
- if config/opamp server defines custom instrumentation it is installed

Hence only 1 thing needs to be configured

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+| `opentelemetry-java-autoinstrumentation` | OpenTelemetry Java agent JAR | `all` / `noarch` |
+| `opentelemetry-nodejs-autoinstrumentation` | OpenTelemetry Node.js auto-instrumentation | `all` / `noarch` |
+| `opentelemetry-dotnet-autoinstrumentation` | OpenTelemetry .NET auto-instrumentation (glibc + musl) | Per-arch (`amd64`, `arm64`) |
+| `opentelemetry` | Metapackage that pulls in the injector and all language packages | `all` / `noarch` |
+
+### Dependency graph
+
+```
+opentelemetry  (metapackage)
+├── opentelemetry-injector1                      (virtual)
```
cpaelzer: Since you explicitly define the kind of dependency below (Provides/Suggests) should this here be explicit as well and mention if you want this to be `Depends(deb)/Requires(rpm)` (if you install the meta package all will be installed, but if you remove any you also pull out the meta package as it can't exist without its childs) or `Recommends` (Installing all, but allowing you to remove one of the dependencies, without pulling out the rest).

Also with Depends this can only exist if all 4 injector/java/nodejs/dotnet are present, with a recommends one could have only one of them.

I do not yet understand enough of opentelemetry to be sure, but I assume this should be
```
├──Depends:         opentelemetry-injector1                      (virtual)
├──Recommends: opentelemetry-java-autoinstrumentation1      (virtual)
├──Recommends: opentelemetry-nodejs-autoinstrumentation1    (virtual)
└──Recommends: opentelemetry-dotnet-autoinstrumentation1    (virtual)
```

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+
+opentelemetry-java-autoinstrumentation
+├── Provides: opentelemetry-java-autoinstrumentation1
+└── Suggests: opentelemetry-injector1
+
+opentelemetry-nodejs-autoinstrumentation
+├── Provides: opentelemetry-nodejs-autoinstrumentation1
+└── Suggests: opentelemetry-injector1
+
+opentelemetry-dotnet-autoinstrumentation
```
cpaelzer, line 63: In case I understood this right you are expecting vendors to provide their own alternative binary packages that can satisfy the same dependency "Provides: opentelemetry-dotnet-autoinstrumentation1" by providing the same files.

I assume and recommend that you still want them to have different binary package names. If they would misunderstand that and use the same binary name the package manager will wreak havoc once their versions race on updates. Going for different names but keeping the same provides seems the right way unless you planned for something else. Could the example please list one of those like:

```
...
somevendorname-opentelemetry-nodejs-autoinstrumentation
├── Provides: opentelemetry-nodejs-autoinstrumentation1
...
```

Furthermore if I understood you right, you intend that those packages shall use the very same file paths on disk. You need to tell the package managers that to avoid conflicts in the dance of package extract<->update<->removal via `Conflicts` [Docs](https://www.debian.org/doc/debian-policy/ch-relationships.html#s-conflicts) so that no two of them can be unpacked at the same time.

```
somevendorname-opentelemetry-nodejs-autoinstrumentation
├── Provides: opentelemetry-nodejs-autoinstrumentation1
├── Conflicts:  opentelemetry-nodejs-autoinstrumentation
...
```


In case I misunderstood feel free to adjust how you meant this to work out, but please consider providing such an example.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+├── opentelemetry-java-autoinstrumentation1      (virtual)
+├── opentelemetry-nodejs-autoinstrumentation1    (virtual)
+└── opentelemetry-dotnet-autoinstrumentation1    (virtual)
+
+opentelemetry-injector
+└── Provides: opentelemetry-injector1
+
+opentelemetry-java-autoinstrumentation
+├── Provides: opentelemetry-java-autoinstrumentation1
+└── Suggests: opentelemetry-injector1
```
cpaelzer, line 57: While the dependency from `opentelemetry  (metapackage)` to the language parts is a Recommends, the other way around is a hard dependency as it can't work with anything but `opentelemetry-injector1` (compatibility). And without `opentelemetry-injector1` it is useless (usability).

I'd therefore have expected `Suggests: opentelemetry-injector1` to be `Depends: opentelemetry-injector1` instead in this and the other examples.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+├── Provides: opentelemetry-nodejs-autoinstrumentation1
+└── Suggests: opentelemetry-injector1
+
+opentelemetry-dotnet-autoinstrumentation
+├── Provides: opentelemetry-dotnet-autoinstrumentation1
+└── Suggests: opentelemetry-injector1
+```
+
+Every dependency in the graph uses a virtual package name rather than a concrete one.
+The trailing `1` is not a package release version — it is the **interface generation number**, following the shared-library SONAME convention (`libssl3`, `libgcc-s1`).
```
cpaelzer, line 69: There is one detail missing to match the established practice - the binary package itself would also be versioned to allow them to be co-installable along transitions or if you actually want to provide more than one at the same time.

You no more need a provides on "the original" in that case.

You can still use the `Depends: opentelemetry-injector1` to ensure the compatible interface is present as well.

For this to make sense `opentelemetry-java-autoinstrumentation1` and a future `opentelemetry-java-autoinstrumentation1` (same for all others) need to be co-installable which is usually done via files&paths to include the version.

I can't tell you if that is the way you want to go, but if you want to fully adopt the model that versioned libraries have that would be the way.

```
opentelemetry  (metapackage)
├── Depends: opentelemetry-injector1
├── Recommends: opentelemetry-java-autoinstrumentation1      (virtual)
├── Recommends: opentelemetry-nodejs-autoinstrumentation1    (virtual)
└── Recommends: opentelemetry-dotnet-autoinstrumentation1    (virtual)

opentelemetry-java-autoinstrumentation1
└── Depends: opentelemetry-injector1
...
somevendor-opentelemetry-java-autoinstrumentation1
├── Provides: opentelemetry-java-autoinstrumentation1
└── Depends: opentelemetry-injector1
```

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+
+```
+opentelemetry  (metapackage)
+├── opentelemetry-injector1                      (virtual)
+├── opentelemetry-java-autoinstrumentation1      (virtual)
+├── opentelemetry-nodejs-autoinstrumentation1    (virtual)
+└── opentelemetry-dotnet-autoinstrumentation1    (virtual)
+
+opentelemetry-injector
+└── Provides: opentelemetry-injector1
```
cpaelzer, line 53: Provides should be versioned (at least in Debian/Ubuntu), because a subtle detail of https://www.debian.org/doc/debian-policy/ch-relationships.html#virtual-packages-provides that is often missed is "... If the Provides field does not specify a version number, it will not satisfy versioned dependencies or violate versioned Conflicts or Breaks...".
Not sure if you want to do this here or later in the document, but I would think about something like the following and a reference to this documentation I linked?

```
opentelemetry-injector 1.2-3
└── Provides: opentelemetry-injector1 (=1.2-3)
```

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+A vendor can ship a replacement package that also provides the same virtual name — combined with `Conflicts`/`Replaces` on the concrete upstream name, the package manager handles the swap transparently.
+If the upstream changes what "being a Java auto-instrumentation provider" means, it bumps to `opentelemetry-java-autoinstrumentation2`; existing vendor packages that still provide `…1` cannot satisfy the new dependency.
+See [Vendor Override](#vendor-override) for the recipe and user experience.
+
+## Filesystem Layout
+
+All paths follow the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html) (FHS).
+
+```
+/usr/lib/opentelemetry/
```
cpaelzer, line 96: Since you took inspiration from versioned libraries they could, and imho should, be even co-installable across versions - at least along transitions between the two. To achieve that you might consider that the base path for this is versioned as well.
Like for now `/usr/lib/opentelemetry/1/` and once there is a v2 `/usr/lib/opentelemetry/2/`.
Or `/usr/lib/opentelemetry1/`  `/usr/lib/opentelemetry2/`.
Or `/usr/lib/opentelemetry/injector1/`  `/usr/lib/opentelemetry/injector2/`.
Or ... up to you.

All package content would then fall into place naturally and v1 and v2 could co-exist if ever needed.
And furthermore one could derive from the file path which interface version it is meant to work with which can be a benefit for admins and debugging.

This isn't strictly needed, but seemed worth to suggest after reading that you want to align following the shared-library SONAME convention.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+├── injector/
+│   └── libotelinject.so
```
cpaelzer, line 98: I do get the point that your ABI actually the "conf.d configuration API" and that you won't expect these to load normally via linkage but via ldpreload - so some/all of this might not apply, but it is worth to be asked. And even if you say no, I'd ask to spell out that you said no and the related reasoning in this document please.

Two things missing here
1. outlining how you want actual SONAMES to work, usually you have the SONAME (what things link to) and then the actual version including minors.
2. it seems not needed now, but at some point the multiarch lobby will get you :-) So how about considering it right now. Architecture dependend binaries like the so file should be in multiarch triplet paths.

```
...
├── injector/
│   └── x86_64-linux-gnu
│           └── libotelinject.so.1
│           └── libotelinject.so.1.2
...


This somewhat differs between https://wiki.debian.org/Multiarch/HOWTO and https://docs.fedoraproject.org/en-US/packaging-guidelines/#_merged_file_system_layout (lib/lib64) but while the exact path differs, it would still be a step in the file path.

File: docs/design/packages-meta-architecture.md [hunk 22]
```diff
@@ -0,0 +1,606 @@
...
+│   └── otel-config.yaml
+├── nodejs/
+│   └── otel-config.yaml
+└── dotnet/
+    └── otel-config.yaml
+
+/usr/share/man/
+├── man8/opentelemetry-injector.8.gz
+└── man1/
+    ├── opentelemetry-java.1.gz
```
cpaelzer: Least important thought, but while I'm reading this let me ask - are those `opentelementry-$language` gonna be user executable commands (section 1 of man pages) or more commands for the system administrator (section 8).

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+
+| Field | DEB | RPM |
+|-------|-----|-----|
+| Architecture | `amd64` or `arm64` | `x86_64` or `aarch64` |
+| Provides | `opentelemetry-injector1` | `opentelemetry-injector1` |
+| Config files | `/etc/opentelemetry/injector` | `/etc/opentelemetry/injector` |
+| Post-install | Appends `libotelinject.so` to `/etc/ld.so.preload` | Same |
+| Pre-uninstall | Removes `libotelinject.so` from `/etc/ld.so.preload` | Same |
+
+The post-install and pre-uninstall scripts must use only POSIX shell builtins (`read`, `case`, shell redirection) to avoid dependencies on `sed` or `grep`.
```
cpaelzer, line 173: Isn't that unnecessary pain? Essential packages (`apt list '?essential'`) can be expected to be around and usable - and that includes sed and grep.

And while to my limited knowledge the very same does not exist in the RPM world, there you can still express it in the metadata as [scriptlet dependencies](https://rpm-software-management.github.io/rpm/manual/more_dependencies.html) like:
```
Requires(pre): coreutils
Requires(post): sed, grep
```

And if you need to be explicit in Debs, [Pre-Depends](https://www.debian.org/doc/debian-policy/ch-relationships.html#binary-dependencies-depends-recommends-suggests-enhances-pre-depends) is there as well (but best is to avoid that - essentials are usually enough).

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+
+#### Injector interface versioning
+
+The injector declares `Provides: opentelemetry-injector1`.
+The trailing `1` is not the package release version — it is the **generation number of the conf.d configuration API**, following the Debian shared-library naming convention (`libssl3`, `libgcc-s1`, `libpng16-16`).
```
cpaelzer, line 181: Worth a link to https://www.debian.org/doc/debian-policy/ch-sharedlibs.html#shared-libraries ?

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+
+The metapackage depends on `opentelemetry-injector1` (hard dependency) to ensure the installed injector is compatible with the language packages it pulls in.
+Language packages only suggest `opentelemetry-injector1` (DEB `Suggests`; omitted on RPM which has no equivalent) since they are useful on their own without the injector.
+This decouples the API contract from the package's release cadence:
+
+- **Today:** the injector provides `opentelemetry-injector1`. The metapackage depends on it. Everything resolves.
+- **If a future injector release breaks the conf.d contract:** that release stops providing `opentelemetry-injector1` and starts providing `opentelemetry-injector2`. The metapackage still depends on `opentelemetry-injector1`, so the package manager blocks the injector upgrade until the metapackage is also updated.
```
cpaelzer, line 188: "package manager blocks the injector upgrade until the metapackage is also updated" ... sadly ... or it considers removing the metapackage as an alternative solution out of this if a user asks for the new version of the injector :-)

Because nothing explicitly depends on the meta-package that will be an option for the dependency solver. You may consider to define the injector to `Depends` on the meta package which would block that from ever happening.

cpaelzer: But then we need to consider that we might get close to ask what is the point of the meta package at all. You could have the Recommends to the opentelemetry-java-autoinstrumentation1, opentelemetry-nodejs-autoinstrumentation1, ... right in opentelemetry-injector. And if it is only about the name, you may just add `Provides: opentelemetry" to `opentelemetry-injector`.
Up to you, but it might remove one of the complexities from this equation

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+**Interface versioning (`opentelemetry-injector1`).**
+Tracks generation 1 of the injector's `conf.d/` configuration API.
+The metapackage depends on this virtual name to ensure the installed injector is compatible with the language packages it pulls in.
+If a future injector release breaks the conf.d contract, it bumps to `opentelemetry-injector2`; the package manager blocks incompatible combinations automatically.
+See [Injector interface versioning](#injector-interface-versioning) for the full upgrade scenario.
+
+**Swappable alternatives with interface versioning (`opentelemetry-java-autoinstrumentation1`, etc.).**
+Tracks generation 1 of the contract between the injector and each language's auto-instrumentation provider — the conf.d key names, the file layout under `/usr/lib/opentelemetry/<language>/`, and the `otel-config.yaml` structure.
+The metapackage depends on these virtual names.
+A vendor can ship a replacement package that also provides the same virtual name — combined with `Conflicts`/`Replaces` on the concrete upstream name, the package manager handles the swap transparently.
```
cpaelzer:
```suggestion
A vendor can ship a replacement package with a different name that also provides the same virtual name — combined with `Conflicts`/`Replaces` on the concrete upstream name, the package manager handles the swap transparently.
```
In case you follow my argument above, updating this to reduce misunderstanding should be considered

docs/design/packages-meta-architecture.md
cpaelzer: Many build environments do not allow to download at build time - rightfully so to have a chance at controlling what the content is and to be reproducible. But it makes sense when you split into "source build" and "binary build".


```suggestion
Maintainer or automation downloads the upstream [OpenTelemetry Java agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) JAR at source build time and packages it. Later builds and re-builds of the same source version will consistently build the very same content. Updating to new JAR content requires to create a new source upload.
```

cpaelzer: as with most of my comments, in case you agree here the same applies to the other build variants below

docs/design/packages-meta-architecture.md
cpaelzer: > "Downloads the ... binaries at build time"

This sounds risky - even with my suggested clarification above of this build here being "source build" while a later upload to a Distro build service/Archive then does run the binary build to create the final packages out of them

The source upload to distributions is meant to be that, source - that can be inspected and patched.

If it can be avoided to upload prebuilt binaries, please do so and build them at the actual build stage after upload.

If it can't be avoided it is possible, but puts quite some restrictions on them in terms of maintainability, what can depend on them and if they are available to more security concerned setups that often disable the repositories containing those.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+
+The post-install and pre-uninstall scripts must use only POSIX shell builtins (`read`, `case`, shell redirection) to avoid dependencies on `sed` or `grep`.
+
+#### Injector interface versioning
+
+The injector declares `Provides: opentelemetry-injector1`.
+The trailing `1` is not the package release version — it is the **generation number of the conf.d configuration API**, following the Debian shared-library naming convention (`libssl3`, `libgcc-s1`, `libpng16-16`).
+
+The metapackage depends on `opentelemetry-injector1` (hard dependency) to ensure the installed injector is compatible with the language packages it pulls in.
+Language packages only suggest `opentelemetry-injector1` (DEB `Suggests`; omitted on RPM which has no equivalent) since they are useful on their own without the injector.
```
cpaelzer: "omitted on RPM which has no equivalent"

Well, I'm not the RPM expert - but while it is true that at the level of RPMs that is ignored "Weak dependencies" in https://rpm-software-management.github.io/rpm/manual/dependencies.html suggests that "dependency solvers (might use it) to make choices about what package to install". Furthermore https://docs.fedoraproject.org/en-US/packaging-guidelines/WeakDependencies/

docs/design/packages-meta-architecture.md
cpaelzer: In case you agree, many places like "Virtual package names" throughout the doc needs to be adapted as well

docs/design/packages-meta-architecture.md
cpaelzer: Seeing the FPM examples below you might only be interested to build your own and never get directly into a Distribution for yum/dnf/apt install without configuring extra repositories - in that case you might ignore the comment about please uploading source and not binaries.

But if the final goal is to get directly into Distributions some day, then it is worth to be considered as the rules are quite similar for most of them.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+apt install opentelemetry
+```
+
+Installs the metapackage and all upstream language packages.
+Every `opentelemetry-<language>-autoinstrumentation1` dependency is satisfied by the corresponding concrete upstream package.
+
+#### Fresh install with a vendor package
+
+```sh
+apt install opentelemetry acme-java-autoinstrumentation
```
cpaelzer, line 494: I like that I see you did in some places already consider "somevendor-java-autoinstrumentation" \o/

Furthermore, if you follow my suggestion to avoid dependency managers to pull out the base package ((https://github.com/open-telemetry/opentelemetry-packaging/pull/10/changes#r3285886914) then this line would just be `apt install acme-java-autoinstrumentation` which is nice as installing and switching then would be the same (one less variation).

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+`dpkg` removes the vendor package (symmetric `Conflicts`/`Replaces` if the vendor chose to declare them, or the user runs `apt remove acme-java-autoinstrumentation` first) and installs upstream.
+
+## Current POC Gaps
+
+The [POC](https://github.com/open-telemetry/opentelemetry-injector/pull/239) in the OpenTelemetry Injector repository implements most of the architecture above but has the following gaps relative to the proposed design:
+
+### 1. No interface version on the injector
+
+The injector package does not declare `Provides: opentelemetry-injector1`.
+There is no mechanism to prevent an incompatible injector upgrade from breaking installed language packages.
```
cpaelzer, line 529: If you follow my suggestions above (https://github.com/open-telemetry/opentelemetry-packaging/pull/10/changes#r3285886914) - then you get this for free.

The `opentelemetry-$language-autoinstrumentation` that needs v1 would Depend on `opentelemetry-injector1` and thereby a newer `opentelemetry-injector` can only be upgraded in a transaction that also updates the languages (or removes them). Both leaving the system in a better state than just becoming incompatible.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+All consumers depend on the concrete package name (`opentelemetry-injector`) instead of a versioned interface name.
+
+### 2. No interface version on language packages
+
+Language packages do not declare `Provides` with a versioned virtual name (e.g., `opentelemetry-java-autoinstrumentation1`).
```
cpaelzer, line 534: Again, the hard dependency from them to the interface ((https://github.com/open-telemetry/opentelemetry-packaging/pull/10/changes#r3285886914)) would solve that.

docs/design/packages-meta-architecture.md
```diff
@@ -0,0 +1,606 @@
...
+| `/usr/share/doc/opentelemetry-injector/` | Documentation and copyright |
+
+#### Package metadata
+
+| Field | DEB | RPM |
+|-------|-----|-----|
+| Architecture | `amd64` or `arm64` | `x86_64` or `aarch64` |
+| Provides | `opentelemetry-injector1` | `opentelemetry-injector1` |
+| Config files | `/etc/opentelemetry/injector` | `/etc/opentelemetry/injector` |
+| Post-install | Appends `libotelinject.so` to `/etc/ld.so.preload` | Same |
```
cpaelzer, line 170: Well, I wish I'd have a better suggestion than preload, but I don't.

Still - it might be worth to point to prior are here or somewhere else.
That would help to a) clarify that you are not the only outlier and b) help to not re-implement everything from scratch

Suggesting: https://salsa.debian.org/pkg-security-team/snoopy/-/blob/debian/master/debian/snoopy.postinst.in https://salsa.debian.org/pkg-security-team/snoopy/-/blob/debian/master/debian/snoopy.prerm

We are a decade too late to benefit from https://launchpad.net/ubuntu/+source/ld.so.preload-manager/+publishinghistory :-P

cpaelzer (commented):
Hi,
sorry it ended up being so many comments, but since I come with no context
due to being asked to have a look out of the blue I have only a few "oh you
should change ..." but many "did you mean a or maybe consider b" (like the
comments about how much do you want to be like lib SONAMES) which you
have to decide and then act or ignore. I have mentioned each topic once,
but if you adapt e.g. the dependency graph for one of my comments there might
be chances that you also need to change the pakcage metadata section. But
bringin up the same topic many times would not help after I've seen how many
things I raised on the first few lines.

I hope that it helps you to get this forward, after all even those you
consciously say "no, we meant it to be that way" will be good to be spelled out :-)

Finally: Humans are limited, and so is reviewing a high level spec.
At some point in this evolution you will have first experimental RPM and Deb
builds and I highly recommend to then run lintian in pedantic mode for debs
and for rpms rpmlint/rpminspec/fedora-review (others would know better which to
pick) and be open to adjust the packaging but also the design with all they
throw at you. I'm essentially asking you to plan for that feedback loop
to happen even after we think "yep, seems reasonable" here.


P.S. It was a bit hard to read a spec, to then find "Current POC Gaps" into
"Required Changes to the POC". I'm not sure if I should try to rewrite all I
said :-/ But after a short check I found that most still applies, I can only assume that at some point you will roll up all that is down there plus all the outcome of the discussion into a single description - right?

docs/design/packages-meta-architecture.md
mmanciop: The `opentelemetry` metapackage is intended to depend on the injector and all of the language packages. It is intended to give a "catch-all" experience, for when the end user does not know upfront which languages to monitor, and wants the widest coverage.

cpaelzer: Sure, but a Recommends instead of a Depends would still achieve "They do not know what they want and use the catch all". But in addition my suggestion would allow later:
They now know and do not want to have opentelemetry-dotnet-autoinstrumentation1, with `Recommends` they can remove it, while with `Depends` they tear out all of it

docs/design/packages-meta-architecture.md
thompson-tomo: It seems that this has blended into the other comment, so let's circle back to the original comment. Are we really needing to recommend a dedicated file per language? Is that not introducing chance of divergence in config ie resource attributes? Could a single opamp supervisor on a system handle deploying the multiple config files of the same type to different paths? Will the injector inject the correct config file path if there is multiple?

mmanciop: > Are we really needing to recommend a dedicated file per language?

Yes, different languages need different instrumentation opt-in/opt-out lists

> Is that not introducing chance of divergence in config ie resource attributes?

It may, but the long-term vision is to have a OpenTelemetry Collector package included in the overall picture and then resource detection would be done there.

> Could a single opamp supervisor on a system handle deploying the multiple config files of the same type to different paths?

I don't know.

> Will the injector inject the correct config file path if there is multiple?

Yes, the one config file per language.

docs/design/packages-meta-architecture.md
mmanciop: At the moment, OpAMP is simply not part of the vision for these packages. OpAMP is a topic that never came up in the definition of the [project](https://github.com/open-telemetry/community/pull/3252). If somebody from @open-telemetry/opamp-go-maintainers and @open-telemetry/opamp-java-maintainers would like to barge in and help out, it is something I'll gladly engage with. But having one file how all language _is not sound_ with respect to instrumentation opt-in and opt-out, which is fundamental when dealing with autoinstrumentations where the user has not upfront control of which are injected.

docs/design/packages-meta-architecture.md
mmanciop: @Kielek can you help us reducing the duplication?

docs/design/packages-meta-architecture.md
mmanciop: I think here there is some confusion about the **package interface version** (the `1` in `opentelemetry-dotnet-autoinstrumentation1`) and the dependencies of the `opentelemetry-dotnet-autoinstrumentation` (notice, no `1` package), which will then have its own **package version** (e.g., v1.2.3) and dependencies / conflicts with the actual .NET packages carrying the runtimes.

docs/design/packages-meta-architecture.md
mmanciop: Done in a1f0f67

docs/design/packages-meta-architecture.md
mmanciop: Done in e72c977

docs/design/packages-meta-architecture.md
mmanciop: Thanks for the pointer to `snoopy` — it's great prior art for the `/etc/ld.so.preload` management pattern.

Regarding the POSIX-builtins-only requirement: while sed and grep are indeed Essential: yes in Debian, the picture is different on RPM-based distros. AFAICT `sed` and `grep` are not in Fedora's or RHEL's mandatory `@core` group — they're pulled in transitively but not formally guaranteed. Minimal and container images (e.g. UBI Micro) may not have them. The RPM docs explicitly state that non-interpreter scriptlet dependencies must be manually declared (`Requires(post): sed`, etc.).                                                                      

We'd rather keep the scripts portable using only POSIX shell builtins and avoid the extra dependency declarations on the RPM side. The `snoopy` Debian package does use `grep`, but it also only targets DEB — for maintainability, we prefer a single set of scripts for both ecosystems.

docs/design/packages-meta-architecture.md
mmanciop: Clarified in 37313af

docs/design/packages-meta-architecture.md
mmanciop: Man 8 it is, see 792c2cf

docs/design/packages-meta-architecture.md
mmanciop: Thanks! Added in 07f6718

docs/design/packages-meta-architecture.md
Kielek: I think that you can mimic operator solution

It downloads both versions, but keep _managed_  and _unamaged_ libs from glibc package and adds there only _unamaged_ part for musl

Ref: https://github.com/open-telemetry/opentelemetry-operator/blob/7531991fa87143ee584e5a993f63d581f5e0fe74/autoinstrumentation/dotnet/Dockerfile#L26-L29

docs/design/packages-meta-architecture.md
mmanciop: Clarified in 66aa7ed

docs/design/packages-meta-architecture.md
mmanciop: Updated the language in 79d22d4, we also need to be able to track it in the source of the package to have (somewhat) reproducible builds

docs/design/packages-meta-architecture.md
mmanciop: Great catch, done in 42c8b2a

docs/design/packages-meta-architecture.md
mmanciop: Having the injector package providing the metapackage name will be detrimental in a later phase when we bring the [OpenTelemetry eBPF Instrumentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation) (OBI) and its system package "into the fold".

It seems to me the switch in 42c8b2a for the metadapackage from `Depends` to `Recommends` wrt language packages takes care of it pretty much.

docs/design/packages-meta-architecture.md
mmanciop: Keeping `Suggests` here. The language packages are genuinely useful without the injector — for example, the Java agent JAR can be used directly via `-javaagent:/usr/lib/opentelemetry/java/opentelemetry-javaagent.jar`. The Node.js instrumentation can be loaded via `NODE_OPTIONS`. Similar environment-activation exists for all the other languages supported by the injector, and it is indeed **the** requirement for a language to be supported by the injector. In other words: the injector automates the activation, but it's not the only way to use these packages.

The metapackage already ensures co-installation for the standard "install everything" use case via `Depends: opentelemetry-injector1`, so users who go through `apt install opentelemetry` always get both.

docs/design/packages-meta-architecture.md
mmanciop: Indeed, per Debian policy, versioned `Provides` would allow consumers to express versioned dependencies in the future. However, we haven't defined the package versioning scheme yet (tracked in #11), and no current consumer requires a versioned dependency. Adding a version to `Provides` now would be mechanical but potentially misleading until the versioning scheme is stable.

30f6023 adds a note to the doc deferring versioned Provides until #11 is resolved. Once the versioning scheme is decided, adding `(= X.Y.Z)` to every `Provides` is a backwards-compatible, one-line change per package.

docs/design/packages-meta-architecture.md
mmanciop: Co-installable versioned packages work well for shared libraries where multiple ABIs genuinely coexist. However, in our case the injector hooks into `/etc/ld.so.preload` and only one version can be active system-wide. Since the injector can only speak one interface generation at a time, there's no scenario where both gen 1 and gen 2 injector packages would be active simultaneously.

For language packages, the argument is different: the end user may wish for multiple interface generations of language packages to be installable in parallel (one to be used by the injector, the others manually). Unfortunately, that that would require embedding the interface version in the file paths (e.g., `/usr/lib/opentelemetry/java-1/...`), which is _guaranteed_ to confuse the end users, who would see the index suffix as related with the runtime version (e.g., Java v1) instead of the much more obscure package interface version, which is well hidden in package metadata and effectively only a concern of the package manager.

Given the fact that wanting multiple generations of language packages seems like an obscure and unlikely requirement (and yes: this might be a case of "last famous words"), I think it is best skipping the additional confusion it brings, and have later different generations of the same language package to use `Conflicts/Replaces`, and for the package manager to handle the transition atomically.

5164eed documents this.

docs/design/packages-meta-architecture.md
mmanciop: Re SONAME versioning: `libotelinject.so` is loaded via a literal path in `/etc/ld.so.preload`, not linked against by any binary. The dynamic linker's `SONAME` resolution doesn't come into play. A versioned filename would just mean that the post-install script writes `libotelinject.so.1` into `/etc/ld.so.preload` instead of `libotelinject.so`, no functional difference. Interface compatibility is tracked at the package level via `Provides: opentelemetry-injector1`.

Re Multiarch triplet paths: Since the injector is loaded into every process via `/etc/ld.so.preload`, only one architecture's injector can be active system-wide. Co-installing `amd64` and `i386` variants (and we do not plan an `i386` build btw) doesn't make sense, so multiarch paths add complexity without benefit. Besides: only some languages have arch-dependent components, and it is never the full package content.

Both SONAME versioning and Multiarch triplet paths feel unwarranted, and this is documented in d2a0387.

docs/design/packages-meta-architecture.md
cpaelzer: Thanks, yes that works and thanks for adding a reminder for yourself to not forget about it

docs/design/packages-meta-architecture.md
cpaelzer: Sure if you can make them work as one for all go for that, just remember that if you do end up adding any of those that you are -not- supposed to add this as dependency. So where the RPM world wants you to `Requires(post): sed` the Deb world does not want that as dependencies on essential are just bloat.

Feel free to mark this closed

docs/design/packages-meta-architecture.md
thompson-tomo: > Yes, different languages need different instrumentation opt-in/opt-out lists

This is news to me and differs to what the schema/example in the config spec shows https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#experimentalinstrumentation which is that you are allowed to set options for multiple languages in the one file.

> Yes, the one config file per language.

I know it would be possible but reading https://github.com/open-telemetry/opentelemetry-injector/blob/main/README.md#environment-modifications it doesn't appear to be existing functionality.

docs/design/packages-meta-architecture.md
thompson-tomo: > OpAMP is simply not part of the vision for these packages. OpAMP is a topic that never came up in the definition of the project.

I don't know why opamp might have been overlooked however the collector couod be considered as an opamp agent hence deploying that introduces opamp. Anyway, I don't particularly see packaging having much interaction with opamp other than deploying the agent/collector. But the key thing for me is not to present to companies that the only option if they want to deploy otel using their own auto-instrumentation is to build their own system package.

The simplest path to address this would be to add a note mentioning something like:

> open telemetry does offer alternative method/s ie opamp to install packages ie the auto instrumentation agent onto a target system. however these processes are out of scope of the system packaging and it is their responsibility to ensure that the package they installed will be loaded automatically and they can co-exist with the system package.

That way it is clear other pathways are available.

> having one file how all language is not sound with respect to instrumentation opt-in and opt-out, which is fundamental when dealing with autoinstrumentations where the user has not upfront control of which are injected.

The config is already split per language for instrumentation options hence shouldn't be an issue.

docs/design/packages-meta-architecture.md
mmanciop: > This is news to me and differs to what the schema/example in the config spec shows https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#experimentalinstrumentation which is that you are allowed to set options for multiple languages in the one file.

I see, I had missed that development.

> I know it would be possible but reading https://github.com/open-telemetry/opentelemetry-injector/blob/main/README.md#environment-modifications it doesn't appear to be existing functionality.

Indeed it does not yet, but it is an easy addition if we need it (and Antoine and I are maintainers for the Injector too).

This being said, if one single config file across multiple languages is indeed feasible, maybe that would both solve OpAMP issues and reduce overhead for end users to define e.g. common resource attributes across all languages. I do however suspect that the moment we move to one config filer for all languages, somebody will definitely come up with a very legitimate reason to split some languages out in their deployments, so we should not lock ourselves in.

frzifus (commented):
@F-X64 could you please help out with RPM specifics?

docs/design/packages-meta-architecture.md
thompson-tomo: I agree we should not lock ourselves in and make it harder to use other otel technologies.

> I do however suspect that the moment we move to one config filer for all languages, somebody will definitely come up with a very legitimate reason to split some languages out in their deployments

Thinking about it some more would defining more structure to the distribution block in declarative config help for those scenarios that we could say for this auto-instrumentation use these options? Just like we already do for tracers etc.

docs/design/packages-meta-architecture.md
trask: > You will be codeowners of this package configuration.

we will? I see @atoulme listed for Java staffing: https://github.com/open-telemetry/community/blob/main/projects/packaging.md#staffing--help-wanted

docs/design/packages-meta-architecture.md
jack-berg: OH that's a good point. Not needed.

docs/design/packages-meta-architecture.md
jack-berg: > for the same language, SDK and contrib packages usually have different versions schemes.

The package is a new composite artifact which needs a versioning scheme of its own, if no obvious candidate already exists (like java agents).

docs/design/packages-meta-architecture.md
atoulme: Sorry, my mistake. We discussed notifying maintainers of other SIGs and let them know of this, but I'm responsible of this code.

docs/design/packages-meta-architecture.md
atoulme: Sorry, my mistake. We discussed notifying maintainers of other SIGs and let them know of this, but that didn't come with a codeownership assignment.

docs/design/packages-meta-architecture.md
atoulme: Sorry, my mistake. We discussed notifying maintainers of other SIGs and let them know of this, but that didn't come with a codeownership assignment.

docs/design/packages-meta-architecture.md
mmanciop: Done in 1a26f59

docs/design/packages-meta-architecture.md
mmanciop: I did some research into the actual state of the `instrumentation.<language>` sections in the [declarative configuration schema](https://github.com/open-telemetry/opentelemetry-configuration) and their implementation across SDKs.

The schema does define per-language keys under `instrumentation/development` (`java`, `js`, `dotnet`, `python`, `cpp`, etc.), which would in theory allow a single config file to carry language-specific instrumentation settings. However, the actual implementation status tells a different story:

| SDK | Declarative config | `instrumentation.<lang>` section | Actually consumed at runtime |
|-----|-------------------|-------------------------------|------------------------------|
| **Java** | Experimental | Defined & supported | **Yes** |
| **.NET** | Experimental | Marked `not_applicable` in [support matrix](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/language-support-status.md); .NET has its own file-based config implementation that does not consume the shared schema's `instrumentation.dotnet` key | Own impl only |
| **Node.js** | Experimental | Marked `unknown`; not implemented in the [configuration package](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/configuration) — it only handles providers, exporters, and propagators | **No** |
| **Python** | In development | Model [defined](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/_configuration/models.py) but never wired up — `_sdk.py` skips instrumentation entirely | **No** (dead field) |
| **C++** | Stable | Marked `not_applicable` | **No** |

**Summary:** Java is the only SDK that actually reads and acts on the language-specific instrumentation section from the shared schema. .NET has its own parallel implementation. Node.js and Python don't support instrumentation configuration from a YAML file at all today.

Given this, I think per-language config files remain the right approach for the system packages for the foreseeable future. If the SDKs converge on a single shared schema for instrumentation in the future, we can revisit, but we'd be designing for a reality that doesn't exist yet.

I did adopt @jack-berg's suggestion to rename the config files from `otel-config.yaml` to `otel-sdk-config.yaml` to future-proof against potential collector config conflicts and to make a single-file consolidation easier down the road if it becomes viable.

docs/design/packages-meta-architecture.md
mmanciop: Done in a1f0f67

docs/design/packages-meta-architecture.md
mmanciop: Thanks for the correction! RPM does support Suggests via weak dependencies. The doc has been updated accordingly.

docs/design/packages-meta-architecture.md
mmanciop: The [Vendor Override](#vendor-override) section (starting around line 350) covers this in detail — vendor packages use different binary names (e.g., `acme-java-autoinstrumentation`), declare `Provides: opentelemetry-java-autoinstrumentation1` to satisfy the virtual dependency, and `Conflicts`/`Replaces` on the upstream concrete name.

docs/design/packages-meta-architecture.md
mmanciop: Addressed in the "Co-installable interface generations" rejected alternative (line 569). The injector hooks into `/etc/ld.so.preload` and only one version can be active system-wide, so co-installing v1 and v2 paths would leave one inactive. Different generations use `Conflicts`/`Replaces` instead, and the package manager handles the transition atomically.

docs/design/packages-meta-architecture.md
mmanciop: Correct — as you noted in the follow-up, we build with FPM for self-hosted repositories, not for upload to official distro archives. The .NET (and Java, Node.js) artifacts are fetched as pre-built upstream releases at source/build time and packaged as-is. The SBOM metadata (line 319) tracks what's bundled for security auditing purposes.

docs/design/packages-meta-architecture.md
mmanciop: Thanks! As noted in the reply on line 55, we're keeping `Suggests` rather than a hard dependency — language packages are useful standalone without the injector. So the vendor install flow remains `apt install opentelemetry acme-java-autoinstrumentation` as documented in the user experience scenarios.

docs/design/packages-meta-architecture.md
mmanciop: See reply on line 378 — we're keeping `Suggests` per the rationale on line 55.

docs/design/packages-meta-architecture.md
mmanciop: See reply on line 378 — we're keeping `Suggests` per the rationale on line 55.

docs/design/packages-meta-architecture.md
mmanciop: Acknowledged — we'll track the .NET team's supported runtime versions when building the package.

docs/design/packages-meta-architecture.md
mmanciop: Done in a1f0f67.

docs/design/packages-meta-architecture.md
mmanciop: Done in a1f0f67.

docs/design/packages-meta-architecture.md
mmanciop: Thanks for the correction! RPM does support `Suggests` via weak dependencies. The doc has been updated — line 173 now reads:

> Language packages only suggest `opentelemetry-injector1` (DEB `Suggests`; RPM [`Suggests`](https://docs.fedoraproject.org/en-US/packaging-guidelines/WeakDependencies/) via `--rpm-tag`) since they are useful on their own without the injector.

There's also a dedicated appendix section ("RPM `Suggests` via FPM") covering the `--rpm-tag` mechanism and the `createrepo_c` requirement for preserving weak dependencies in repo metadata.

docs/design/packages-meta-architecture.md
mmanciop: Agreed, not needed given the per-language path prefix.

mmanciop:
Ongoing PoC at #18

atoulme (approved):

x1unix (approved):

docs/design/packages-meta-architecture.md
mmanciop: Due to good improvements of the configuration support in SDKs, we went in the end with one configuration  file for all languages

mmanciop: The OpAMP discussion is a bit stuck: https://github.com/open-telemetry/opamp-spec/issues/360

Issue #4, opened 2026-05-13
Investigate package hosting options


mmanciop:
We need a place to host the built packages that can host at least DEB and RPM repositories, and ideally (forward-looking), APK. We have considered (ab)using GitHub pages in the beginning, but it has complications with key and signing (PGP) and the like. 

We need to:
[ ] find an option to host these packages like https://openbuildservice.org/, Canonical's Launchpad, or the like
[ ] be able to point at it via a CNAME like `packages.opentelemetry.io`
[ ] have a good way to manage SSL certificates (ideally, a letsencrypt certificate we can control)
[ ] have a good way to manage (PGP) signing, ideally doing the same as what the OpenTelemetry Collector release builds do

The hosting solution _may_ affect the way we need to build the packages.

x1unix:
# Potential options:

## Open Build Service

[OBS](https://openbuildservice.org/) is a generic system to build and distribute binary packages from sources.

Besides self-hosting, there is a public OBS instance hosted by OpenSuse: https://build.opensuse.org/

### Who is using

The public instance is used by different projects to build packages and host repos:

- [Podman](https://build.opensuse.org/repositories/devel:kubic:libcontainers:stable)
- [Kubernetes](https://build.opensuse.org/repositories/isv:kubernetes:core:stable:v1.33)
  - See: https://kubernetes.io/blog/2023/08/15/pkgs-k8s-io-introduction/
- [ownCloud](https://build.opensuse.org/repositories/isv:ownCloud:desktop)
- Distros: [Alpine Linux](https://build.opensuse.org/repositories/Alpine:Latest), [Arch Linux (community repo)](https://build.opensuse.org/repositories/Arch:Community) and [Alma Linux](https://build.opensuse.org/repositories/AlmaLinux:10).

### Checklist

- ✅ Cloud option is available via OpenSuse OBS instance.
- ✅ Provides runners to build packages for different distros and architectures.
- ✅ Supports repo metadata generation for PPA and DEB repos.
- ✅ GPG signing automation is supported, but [you've to renew keys yourself](https://iriarte.it/homelab/2023/10/07/deprecated-gpg-key-in-openbuild-service.html).
- ❌CNAME is not supported, reverse/edge proxy is required.

### Who is using?

Kubernetes is [using public OBS instance](https://kubernetes.io/blog/2023/08/15/pkgs-k8s-io-introduction/) to build packages and generate repo metadata.

[...]

PR #18, opened 2026-06-19
feat: implement metapackage architecture

### LINKED ISSUE #5: Port packaging over the open-telemetry/opentelemetry-injector

mmanciop:
Implement #10 
Closes #5


cmd/build-packages/main.go
```diff
@@ -0,0 +1,106 @@
...
+	"strings"
+
+	"github.com/open-telemetry/opentelemetry-packaging/packaging/builder"
+)
+
+func main() {
+	log.SetFlags(0)
+
+	version := flag.String("version", "", "Package version (required)")
+	arch := flag.String("arch", "amd64", "Target architecture (amd64 or arm64)")
```
carlosalberto, line 35: Valid value is not verified as `format` is - it does fail though, as whatever you pass here will be appended to the package name, e.g. `libotelinject_WRONGARCH.so`

docs/design/integration-test-plan.md
```diff
@@ -0,0 +1,186 @@
...
+### Vendor replacement tests
+
+These need a build step for the mock `acme-*` package, then use the same lifecycle container pattern.
+
+## Makefile targets
+
+```
+make integration-tests                    # all tests
+make integration-test-metadata            # package metadata + contents
+make integration-test-deb-lifecycle       # DEB install/remove/upgrade/config
```
carlosalberto, line 194: `-lifecycle` targets are not defined (yet?)

docs/design/integration-test-plan.md
```diff
@@ -0,0 +1,186 @@
...
+make integration-test-rpm-lifecycle       # RPM install/remove/upgrade/config
+make integration-test-deb-vendor          # DEB vendor replacement
```
carlosalberto, line 196: Likewise with `-vendor` targets.

docs/design/integration-test-plan.md
```diff
@@ -0,0 +1,186 @@
...
+make integration-test-rpm-vendor          # RPM vendor replacement
+make integration-test-deb-java            # E2E telemetry (existing)
```
carlosalberto: With both rpm/deb I'm getting errors with Docker (haven't tried with Podman), even though I can see containers actually running: 

```sh
Step 4/15 : FROM fedora:41
    dotnet_test.go:20: 
        	Error Trace:	/Users/carlosalberto/development/opentelemetry-packaging/testutil/testutil.go:118
        	            				/Users/carlosalberto/development/opentelemetry-packaging/packaging/tests/rpm/dotnet/dotnet_test.go:20
        	Error:      	Received unexpected error:
        	            	create container: build image: NotFound: content digest sha256:e04278bca37cf98cc645106d1bc4e099db65991c2bc9c086d0664896cebe230c: not found
        	Test:       	TestDotnetAutoInstrumentation
--- FAIL: TestDotnetAutoInstrumentation (36.64s)
FAIL
FAIL	github.com/open-telemetry/opentelemetry-packaging/packaging/tests/rpm/dotnet	38.194s
```

docs/design/integration-test-plan.md
```diff
@@ -0,0 +1,186 @@
...
+make integration-test-deb-nodejs          # E2E telemetry (existing)
+make integration-test-deb-dotnet          # E2E telemetry (existing)
```
carlosalberto: Specifically with `deb-dotnet` I'm getting a slightly different error:

```
W: Invalid 'Date' entry in Release file /var/lib/apt/lists/partial/_local-repo_dists_stable_Release
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package opentelemetry-injector
E: Unable to locate package opentelemetry-dotnet-autoinstrumentation
 ---> Removed intermediate container 0ad1c5131a01
    dotnet_test.go:20: 
        	Error Trace:	/Users/carlosalberto/development/opentelemetry-packaging/testutil/testutil.go:118
        	            				/Users/carlosalberto/development/opentelemetry-packaging/packaging/tests/deb/dotnet/dotnet_test.go:20
        	Error:      	Received unexpected error:
        	            	create container: build image: The command '/bin/sh -c apt-get update && apt-get install -y     opentelemetry-injector     opentelemetry-dotnet-autoinstrumentation' returned a non-zero code: 100
        	Test:       	TestDotnetAutoInstrumentation
--- FAIL: TestDotnetAutoInstrumentation (5.03s)
FAIL
FAIL	github.com/open-telemetry/opentelemetry-packaging/packaging/tests/deb/dotnet	5.437s
FAIL

```

carlosalberto:
I'm not sure _these_ packages' documentation the place to mention this, but when enabling declarative file configuration via `OTEL_EXPERIMENTAL_CONFIG_FILE`, environment variables like `OTEL_SERVICE_NAME` are not consumed directly anymore, but rather interpolated at the configuration file - see the [updated documentation](https://opentelemetry.io/docs/languages/java/configuration/#declarative-configuration).

On a related note, this feature is now stable and it should be appearing as `OTEL_CONFIG_FILE` instead (depends on the implementation status, though, e.g. Java already has this as a stable feature).

cmd/build-packages/main.go
mmanciop: Fixed in 85399b2 (`fix: pin integration-test container build architecture`): `-arch` is now validated the same way `-format` is, failing fast with `error: unknown architecture "WRONGARCH" (expected amd64 or arm64)` instead of producing a broken download URL.

docs/design/integration-test-plan.md
mmanciop: This was an architecture mismatch on Apple Silicon: Docker resolved a different platform manifest mid-build than the one the test requested. Fixed in 85399b2, which pins the container build platform to `linux/$ARCH` and exports `ARCH` from the Makefile so the packages, the local repo, and the test containers all agree on one architecture. The suites were since also restructured into a single base-image matrix asserting via the otel-sink (see the `test(<lang>): unify deb/rpm into one base-image matrix over the sink` commits).

docs/design/integration-test-plan.md
mmanciop: This was actually two separate bugs: The `W: Invalid 'Date' entry` warning is fixed in 2f338ab (`fix: add Date field to generated APT Release file`). The fatal `Unable to locate package` was the same architecture mismatch as your rpm failure: the repo contained amd64 packages while the container resolved to arm64, so apt found no installable candidates — fixed by the `ARCH` alignment in 85399b2.

mmanciop:
> I'm not sure _these_ packages' documentation the place to mention this, but when enabling declarative file configuration via `OTEL_EXPERIMENTAL_CONFIG_FILE`, environment variables like `OTEL_SERVICE_NAME` are not consumed directly anymore, but rather interpolated at the configuration file - see the [updated documentation](https://opentelemetry.io/docs/languages/java/configuration/#declarative-configuration).
> 
> On a related note, this feature is now stable and it should be appearing as `OTEL_CONFIG_FILE` instead (depends on the implementation status, though, e.g. Java already has this as a stable feature).

All sample configs, READMEs, and man pages now reference the stable `OTEL_CONFIG_FILE` and `file_format: "1.0"`: fc900dd (Java, plus agent 2.29.0 so the variable actually works), 99ac4a3 (Node.js), e33dfab (.NET, including the additional `OTEL_EXPERIMENTAL_FILE_BASED_CONFIGURATION_ENABLED` flag it still needs), and b7e48bc (Python, bundling the `file-configuration` deps without which setting the variable crashes the configurator).

On the interpolation point: agreed, and it turns out to be a feature for these packages rather than a caveat. A system-wide declarative file can interpolate the values the injector already injects — `endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT}/v1/traces` plus `headers_list: ${OTEL_EXPORTER_OTLP_HEADERS}` (which accepts the env-var wire format as-is) — so `default_env.conf` remains the single source of endpoint and credentials for both env-var-configured and declaratively configured processes during a migration.

mmanciop:
**Update on the .NET declarative-configuration findings**: .NET is *not* failing to deliver telemetry after all.

File-based configuration in auto-instrumentation 1.15.0 works, but it inverts the enablement default: env-var mode turns all instrumentations on unless excluded, while file-based mode enables **none** until they are listed explicitly:

```yaml
instrumentation/development:
  dotnet:
    traces:
      httpclient: {}
```

With that section (plus `OTEL_EXPERIMENTAL_FILE_BASED_CONFIGURATION_ENABLED=true` and `OTEL_CONFIG_FILE`), an unmodified .NET app instrumented by the packaged agent exported spans end to end via the injector-injected configuration.

The remaining upstream concern is much narrower than previously stated: a config file without that section yields a silently telemetry-free process (exporters run, nothing is produced, no warning points at the missing section), and the default differs both from .NET's own env-var mode and from how Java, Node.js, and Python behave under the same shared file.

Findings doc corrected in a73a878; the packaged sample config at `/etc/opentelemetry/dotnet/otel-config.yaml` now shows the instrumentation section (ffe6417).

docs/design/integration-test-plan.md
mmanciop: Implemented in ac6f324

docs/design/integration-test-plan.md
mmanciop: Implemented in ac6f324

packaging/common/python/README.md
```diff
@@ -0,0 +1,89 @@
...
+```bash
+export OTEL_CONFIG_FILE=/etc/opentelemetry/python/otel-config.yaml
+```
+
+## Safety guards
+
+`sitecustomize.py` performs the following checks at startup before activating instrumentation:
+
+1. **Python version**: Requires Python ≥ 3.10. Older versions are skipped gracefully.
+2. **OTLP protocol**: Requires `OTEL_EXPORTER_OTLP_PROTOCOL` to be set and not `grpc`.
```
MikeGoldsmith: When using declarative config, these env vars are ignored and won't be enforced by the env var.

We should add a guard for when `OTEL_CONFIG_FILE` is set, we check any OTLP exporter is httlp/protobuf.

packaging/common/python/requirements.txt
```diff
@@ -0,0 +1,89 @@
...
+# TODO: publish the pure-Python pyproto exporter chain to PyPI and replace the
+# three git+ requirements below with a single pinned release, e.g.:
+#   opentelemetry-exporter-otlp-pyproto-http==<version>
+#
+# Until published, the exporter and its two pyproto-only dependencies live in a
+# fork branch and pin each other at 1.43.0.dev. They are pure-Python (no
+# google-protobuf C extension) and are installed from source by the build's VCS
+# pass. The build installs git+ requirements with --no-deps, so the exporter's
+# own runtime dependency on `requests` is declared explicitly below.
+opentelemetry-exporter-otlp-pyproto-http @ git+https://github.com/ocelotl/opentelemetry-python.git@no-requests#subdirectory=exporter/opentelemetry-exporter-otlp-pyproto-http
```
MikeGoldsmith: I'm wary of using an external directly fork here. I know we're pushing to donate to the official repos and release an official package.

In the mean time, I suggest we publish a temp package to pypi so at least it's versioned properly.

docs/notes/2026-07-04-declarative-config-findings.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Python (upstream: opentelemetry-python)
+
+- The file configurator should declare its dependencies: without the `file-configuration` extra, setting `OTEL_CONFIG_FILE` crashes with a bare `ModuleNotFoundError: No module named 'yaml'` that does not mention the fix (the follow-up `jsonschema` error does).
+  Bundling the extra works around this in our package.
+- The file configurator hardcodes `opentelemetry.exporter.otlp.proto.http` for the `otlp_http` exporter type; custom exporters (our pure-Python `pyproto` exporter) cannot be selected declaratively because Python has no declarative plugin resolution yet.
+  Until it does, declarative configuration forces `opentelemetry-exporter-otlp-proto-http` — and its protobuf dependency — back into our bundle, undercutting the pyproto swap for declarative-config users.
+- Our own `sitecustomize` dependency-conflict guard makes system-Python activation fragile: every dependency added to the bundle is a new potential conflict with distro packages.
+  Concretely, adding pyyaml made the guard refuse activation on Debian 12 (`PyYAML: required ==6.0.3, found 6.0` from the distro's `python3-yaml`); clean virtualenvs are unaffected.
```
MikeGoldsmith: Can we loosen or update the debian target to make this less fragile?

packaging/common/python/sitecustomize.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def _check_dependency_version_conflict(req_string, version_conflicts):
+    """Check for a dependency version conflict. Accumulates conflicts in version_conflicts (modified in place)."""
+    import importlib.metadata
+    from packaging.requirements import Requirement
+    from packaging.version import Version
+
+    _log_debug("_check_dependency_version_conflict({})".format(req_string))
+    req = Requirement(req_string)
```
MikeGoldsmith: I think we need to wrap in a try/except and warn if an error is raised. Same for Version below.

docs/notes/2026-07-04-declarative-config-findings.md
```diff
@@ -0,0 +1,160 @@
...
+resource:
+  attributes:
+    - name: service.name
+      value: ${DECL_SERVICE:-declarative-injected}
+tracer_provider:
+  processors:
+    - batch:
+        exporter:
+          otlp_http:
+            endpoint: https://ingress.eu-west-1.aws.dash0-dev.com/v1/traces
```
MikeGoldsmith: Probably shouldn't use vendor specific URLs in OTel code/docs.

packaging/common/python/otel-config.yaml
```diff
@@ -0,0 +1,23 @@
+# Copyright The OpenTelemetry Authors
```
MikeGoldsmith: Why are these config files commented out? They will be ignored unless `OTEL_CONFIG_FILE` is set, which would break unless these are uncommented.

PS Would it make sense to the config set once and copied into each language? They are portable across languages.

MikeGoldsmith (commented):
Looks pretty good, thanks @mmanciop. I've left some comments and suggestions.

packaging/common/python/requirements.txt
```diff
@@ -0,0 +1,89 @@
...
+# https://github.com/dash0hq/dash0-operator/blob/main/images/instrumentation/python/requirements.txt
+# with opentelemetry-exporter-otlp-proto-http replaced by opentelemetry-exporter-otlp-pyproto-http
+# (pure-Python protobuf, no google-protobuf C-extension dependency).
+
+opentelemetry-distro==0.64b0
+
+# The file-configuration extra (pyyaml, jsonschema) enables declarative
+# configuration via OTEL_CONFIG_FILE; without it, setting that variable makes
+# the configurator crash with ModuleNotFoundError.
+opentelemetry-sdk[file-configuration]==1.43.0
```
MikeGoldsmith, line 12: FYI - declarative config will be moving to it's own package `opentelemetry-configuration` in a future release:
- https://github.com/open-telemetry/opentelemetry-python/pull/5356

testutil/otelsink/sink.go
```diff
@@ -0,0 +1,254 @@
...
+// TestIDAttribute is the resource attribute key used to attribute telemetry to
+// the test that produced it.
+const TestIDAttribute = "test.id"
+
+// hostGateway is the DNS name testcontainers exposes inside a container for
+// reaching host ports declared via ContainerRequest.HostAccessPorts.
+const hostGateway = "host.testcontainers.internal"
+
+// Sink is a running in-process OTLP receiver. Create one with Start.
+type Sink struct {
```
atoulme, line 59: I think it's ok if packaging maintains this for a while, though I'd be open to move it to https://github.com/open-telemetry/opentelemetry-go-build-tools or https://github.com/open-telemetry/opentelemetry-collector-contrib/ just like telemetrygen or golden.

.github/workflows/publish-repos.yml
```diff
@@ -0,0 +1,108 @@
...
+            fedora:latest /scripts/generate-rpm-repo.sh /repo
+
+          # Generate .repo file
+          REPO_URL="https://${{ github.repository_owner }}.github.io/${{ github.event.repository.name }}"
+          cat > repo/rpm/opentelemetry.repo <<EOF
+          [opentelemetry]
+          name=OpenTelemetry Auto-Instrumentation System Packages
+          baseurl=${REPO_URL}/rpm/packages
+          enabled=1
+          gpgcheck=0
```
atoulme, line 94: a note that we will need to follow up on gpgcheck

packaging/repo/index.html
```diff
@@ -0,0 +1,117 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>OpenTelemetry Auto-Instrumentation System Packages</title>
```
atoulme: not really important, but maybe worth making this more general
```suggestion
  <title>OpenTelemetry System Packages</title>
```

packaging/repo/index.html
```diff
@@ -0,0 +1,117 @@
...
+    h1, h2 { border-bottom: 1px solid var(--border-color); padding-bottom: 0.5rem; }
+    h2 { margin-top: 2rem; }
+    a { color: var(--link-color); }
+    .version { color: #666; font-size: 0.9em; }
+    ul { padding-left: 1.5rem; }
+    li { margin: 0.5rem 0; }
+  </style>
+</head>
+<body>
+  <h1>OpenTelemetry Auto-Instrumentation System Packages</h1>
```
atoulme:
```suggestion
  <h1>OpenTelemetry System Packages</h1>
```

packaging/repo/index.html
```diff
@@ -0,0 +1,117 @@
...
+  <pre><code># Add the repository
+echo "deb [trusted=yes] @REPO_URL@/debian stable main" | sudo tee /etc/apt/sources.list.d/opentelemetry.list
+
+# Update and install
+sudo apt update
+sudo apt install opentelemetry</code></pre>
+
+  <h2>Quick Install (RHEL/Fedora/Amazon Linux)</h2>
+  <pre><code># Add the repository
+cat &lt;&lt;EOF | sudo tee /etc/yum.repos.d/opentelemetry.repo
```
atoulme, line 73: alternatively can they download the .repo file from the website?

packaging/repo/index.html
```diff
@@ -0,0 +1,117 @@
...
+# Install
+sudo dnf install opentelemetry</code></pre>
+
+  <h2>Available Packages</h2>
+  <ul>
+    <li><code>opentelemetry</code> - Metapackage that installs all components</li>
+    <li><code>opentelemetry-injector</code> - LD_PRELOAD-based auto-instrumentation injector</li>
+    <li><code>opentelemetry-java-autoinstrumentation</code> - Java agent for automatic instrumentation</li>
+    <li><code>opentelemetry-nodejs-autoinstrumentation</code> - Node.js agent for automatic instrumentation</li>
+    <li><code>opentelemetry-dotnet-autoinstrumentation</code> - .NET agent for automatic instrumentation</li>
```
atoulme, line 90: you added python since - maybe can be added, but not a blocker here

testutil/otelsink/wait.go
```diff
@@ -0,0 +1,68 @@
...
+// schedule, so telemetry arrives asynchronously.
+func (s *Sink) WaitForTraces(t *testing.T, timeout time.Duration, cond func(*Traces) bool) *Traces {
+	t.Helper()
+	deadline := time.Now().Add(timeout)
+	for {
+		tr := s.Traces(t)
+		if cond(tr) {
+			return tr
+		}
+		if time.Now().After(deadline) {
```
atoulme, line 26: maybe use require.EventuallyWithT for all those

CONTRIBUTING.md
```diff
@@ -0,0 +1,187 @@
+# Contributing to OpenTelemetry Packaging
+
+## Prerequisites
+
+- **Go 1.22+** (package builds and tests are pure Go)
```
atoulme:
```suggestion
- **Go 1.26+** (package builds and tests are pure Go)
```

packaging/builder/download.go
```diff
@@ -0,0 +1,535 @@
...
+	switch cfg.Arch {
+	case "amd64":
+		dotnetArch = "x64"
+	case "arm64":
+		dotnetArch = "arm64"
+	default:
+		return fmt.Errorf("unsupported architecture for .NET: %s", cfg.Arch)
+	}
+
+	baseURL := "https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases/download"
```
atoulme, line 183: a follow-up will be to read and check checksums and/or signatures (and maybe if possible, embed them in the package?)

docs/notes/2026-07-04-declarative-config-findings.md
```diff
@@ -0,0 +1,160 @@
+# Declarative configuration through the packages and the injector
+
+Date: 2026-07-04.
+Environment: Debian 12 (arm64) Lima VM, packages from this repository installed, injector active system-wide via `/etc/ld.so.preload`.
+
+Goal: switch a whole host to [declarative configuration](https://opentelemetry.io/docs/languages/sdk-configuration/declarative-configuration/) using only the mechanisms our packages provide.
+The injector injects `OTEL_CONFIG_FILE` (from `default_env.conf`) into every process, the packaged agents pick it up, and completely unmodified applications export to Dash0 as configured by a single system-wide YAML file.
```
atoulme:
```suggestion
The injector injects `OTEL_CONFIG_FILE` (from `default_env.conf`) into every process, the packaged agents pick it up, and completely unmodified applications export to an observability backend as configured by a single system-wide YAML file.
```

docs/notes/2026-07-04-declarative-config-findings.md
atoulme: some other mentions in this file to clean up

packaging/common/python/requirements.txt
```diff
@@ -0,0 +1,89 @@
+# Modelled after
+# https://github.com/dash0hq/dash0-operator/blob/main/images/instrumentation/python/requirements.txt
```
atoulme, line 2: a follow-up we might want to do proper attribution in a NOTICE file. Not a blocker.

packaging/common/scripts/preuninstall-injector.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/sh
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Pre-uninstall script for opentelemetry-injector package.
+# Removes the injector library from /etc/ld.so.preload.
```
atoulme, line 7: maybe a followup to add tests checking uninstall performs as expected.

atoulme (approved):
this is good to merge imo. I will conduct one more test of forking the repository, making a release and publishing a repo on my fork if I can, but overall this is a great way to get us started. I have peppered the PR with follow-ups that I would like to open as issues.
Maybe the only blocking item would be to remove mentions of Dash0 from the notes doc, but I don't mind if that's done in a follow-up.

testutil/otelsink/sink.go
mmanciop: I think this can stay in packaging until we have a second use for it. I think the Injector could benefit too tbh. WDYT @atoulme ?

.github/workflows/publish-repos.yml
mmanciop: It's part of https://github.com/open-telemetry/opentelemetry-packaging/issues/4

packaging/repo/index.html
mmanciop: Meh I do not believe it is good ergonomics.

packaging/builder/download.go
mmanciop: Absolutely, see #21

testutil/otelsink/wait.go
mmanciop: I would rather keep the bespoke helpers, for three reasons:

1. Predicate-style conditions fit the sink's API. The sink exposes functional filters (`tr.WithKind(...).Len() > 0`, the generic `otelsink.NonEmpty`), so wait conditions are one-line booleans. `EventuallyWithT` wants the condition expressed as `assert.*` calls against a `CollectT`, which is a a bit hawkward for "has at least one span matching X".
3. Tailored timeout diagnostics. On timeout the helpers report the sink's `test.id` and how much telemetry was seen (timed out after 30s waiting for traces (`test.id=…`, 0 spans seen)), which immediately separates "nothing arrived — wiring/injector broken" from "telemetry arrived but with the wrong shape". `EventuallyWithT` reports the last tick's collected assertion failures, which only carries that information if every condition is rewritten as assertions.

The cost is ~50 lines of dependency-free code shared by all suites, which seems a fair trade.

atoulme: ok

testutil/otelsink/sink.go
atoulme: Exactly. I'd just say a year from now, I'd hope this has moved out because it's used in a couple places, and it becomes a nice commodity.

packaging/common/python/README.md
mmanciop: Fixed in 48e8eaf

packaging/common/python/requirements.txt
mmanciop: For now, it lives in this repo, see in 7546815

docs/notes/2026-07-04-declarative-config-findings.md
mmanciop: Fixed in 48e8eaf

packaging/common/python/sitecustomize.py
mmanciop: Fixed in 05e8781

docs/notes/2026-07-04-declarative-config-findings.md
mmanciop: Fixed in a5040ec

packaging/common/python/otel-config.yaml
mmanciop: Fixed in 77fe70d

packaging/repo/index.html
mmanciop: Fixed in 1a6a03d

packaging/common/python/requirements.txt
mmanciop: Fixed in f99fbe6

packaging/common/scripts/preuninstall-injector.sh
mmanciop: Fixed in 272dd9f

atoulme (approved):

MikeGoldsmith (approved):
Looks good, thanks @mmanciop 👍🏻

packaging/common/scripts/preuninstall-injector.sh
```diff
@@ -0,0 +1,119 @@
...
+# On upgrade, keep the preload entry: dpkg invokes prerm with "upgrade", rpm
+# invokes %preun with the count of remaining package instances ("1"). On RPM
+# the old version's %preun runs AFTER the new version's %post, so cleaning up
+# here would strip the entry the new version just configured. Only a real
+# removal (dpkg "remove", rpm "0") cleans up.
+# Captured before the field-splitting `set --` below can clobber it.
+action="${1:-}"
+
+case "$action" in
+    upgrade|1)
```
x1unix: Might be relevant to capture upgrade error states as well:

```suggestion
    `upgrade*|deconfigure*|failed-upgrade*|1)
```

CLAUDE.md
douglascamata, line 1: Might want to symlink this to `AGENTS.md` as well. Or vice-versa.

CONTRIBUTING.md
```diff
@@ -0,0 +1,296 @@
...
+2. Add a version pin file `packaging/common/<lang>/release.txt`.
+3. Add a download function in `packaging/builder/download.go`.
+4. Add a component definition in `packaging/builder/components.go` (follow the pattern of `javaInfo`).
+5. Register the component in the `AllComponents` slice.
+6. Add metadata tests in `packaging/tests/metadata/metadata_test.go`.
+7. Add a matrix integration test in `packaging/tests/<lang>/`: one `<lang>_test.go` with a `{deb, rpm} × base image` matrix, plus `Dockerfile.deb` and `Dockerfile.rpm` parameterized by `BASE_IMAGE`. Assert on the exported telemetry via `testutil/otelsink`.
+8. Add the language to the `languages` list in `packaging/tests/lifecycle/lifecycle_test.go`, so the install scenarios cover it.
+9. Add the language to the `lang` matrix in `.github/workflows/build.yml`.
+
+## Cutting a release
```
douglascamata: I'm not highly confident in having release instructions in an AGENTs.md-style file. Agents can accidentally (or not) go wild sometimes. It might be safer to leave these instructions in the README.md for humans to use and follow or just make the whole release process by a Github Action that maintainers can trigger. Just a suggestion though.

douglascamata (commented):
Did an initial pass and it seems okay to me. I need more time to try this out in a VM, but so far so good. Just a few super minor and non-blocking suggestions.

.github/workflows/build.yml
```diff
@@ -0,0 +1,138 @@
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Run sitecustomize.py under every supported Python interpreter generation
+        run: make integration-test-sitecustomize
+
+  build-packages:
+    runs-on: ${{ matrix.ARCH == 'amd64' && 'ubuntu-24.04' || 'ubuntu-24.04-arm' }}
+    strategy:
+      matrix:
+        SYS_PACKAGE: [ "deb", "rpm" ]
```
x1unix, line 55: For `rpm` case, each `SYS_PACKAGE=rpm ARCH=*` matrix job will produce a package with identical `noarch` name.

It happens, because [`packaging/builder/components.go:142`](https://github.com/mmanciop/opentelemetry-packaging/blob/packages-poc/packaging/builder/components.go#L142) overrides arch value, which is used to generate RPM package file name:

```
	if format == "rpm" {
		info.Arch = "noarch"
	}
```

Idk whether this is a concern, at the worst case it will end up building the same package twice.

packaging/common/scripts/preuninstall-injector.sh
mmanciop: Fixed in f898223, with a deb lifecycle test in 80fd2e4 asserting the preload entry survives `prerm` `upgrade`/`failed-upgrade`/`deconfigure`.

.github/workflows/build.yml
mmanciop: Verified: the metapackage and the Java/Node.js agents are noarch/`_all`, so every ARCH matrix job emits them under the same filename, and the release job globbed `artifacts/**/*` across both per-arch dirs — two assets with the same basename, which would fail the first tagged release. Fixed in 25965e8 by staging all artifacts into one directory (coalescing the per-arch duplicates) before uploading.

CLAUDE.md
mmanciop: Not a symlink (https://code.claude.com/docs/en/memory#agents-md), but yes, will fix

mmanciop: Made AGENTS.md canonical in 82321f9, keeping a one-line CLAUDE.md (`@AGENTS.md` import). Claude Code doesn't read AGENTS.md out of the box, and the import is preferred over a symlink for cross-platform (Windows) compatibility; every other agent reads AGENTS.md natively.

CONTRIBUTING.md
mmanciop: I think the convention in OTel is a `RELEASING.md` file, so I'll move it there and leave a reference in `CONTRIBUTING.md`

mmanciop: Fixed in fb81315 — moved the release process out of CONTRIBUTING.md into a dedicated RELEASING.md (with a pointer left behind). Worth noting the safety angle too: the release can't be triggered from docs — `publish-release` is gated to a `v*` tag push on the upstream repo, so it needs maintainer push access regardless of where the steps are written.

x1unix (approved):

PR #62, opened 2026-08-05
docs: clarify declarative config as reference files and establish comment conventions

### LINKED ISSUE #60: Create language-specific reference declarative configuration files
## Background

Each language package currently installs the same file — `packaging/common/otel-config.yaml` — at a language-specific path (`/etc/opentelemetry/<language>/otel-config.yaml`).
That file is designed as a reference: users are expected to copy or adapt it to a central location and activate it via `OTEL_CONFIG_FILE` in `default_env.conf`.

## Problem

Because all four copies are identical, a user who installs only `opentelemetry-java` gets a reference file that contains a `.NET`-specific `instrumentation/development` section that does not apply to them and may be confusing.
Conversely, a user who installs only `opentelemetry-dotnet` gets a file with no guidance on which `.NET` instrumentations to list — that section currently ships with only `aspnetcore` and `httpclient` as examples, with no indication that the list needs to be extended for their specific application.

More broadly, a single shared file cannot carry language-specific annotations, defaults, or instrumentation listings without cluttering the reference for every other language.

## Proposal

Replace the single shared source with four language-specific reference files under `packaging/common/{java,nodejs,dotnet,python}/otel-config.yaml`.
Each file should:

- Include only the SDK configuration sections relevant to that language.
- Carry language-specific comments where behaviour differs (the `.NET` `instrumentation/development` section is the clearest example).
- Serve as a self-contained starting point for a user of that language who wants to adopt declarative configuration.

The common structural content (resource attributes, exporter configuration via environment-variable interpolation) can be kept consistent across the four files without needing to be a single shared source.

mmanciop:
## Summary

- The `otel-config.yaml` shipped by each language package is a reference for users to copy and adapt to a central location, not a file to activate in-place at its per-language path. Updates `README.md`, `otel-config.yaml`, and `default_env.conf` to reflect this, with a concrete central-path example (`/etc/opentelemetry/config.yaml`).
- Establishes a comment convention across all `.conf` and config YAML files under `packaging/common/`: `##` for prose/explanatory comments, `#` for commented-out options users can activate by removing the prefix.
- Applies the convention consistently across all six files under `packaging/common/`.

## Related

- Closes #60 (pre-condition: language-specific reference files don't exist yet, but this PR clarifies the intent in the interim)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -49,6 +49,7 @@ make clean                     # Remove build/
 - Man pages use section 8 (system administration).
 - All config files are `config|noreplace` type in nfpm (preserved on upgrade).
 - nfpm `Info` structs must set `Platform: "linux"` explicitly — omitting it causes malformed `Architecture` fields in DEB packages.
+- In configuration files (`.conf`, `.yaml` shipped under `packaging/common/`), use `##` for prose/explanatory comments and `#` for commented-out configuration entries that users can activate by removing the `#` prefix.
 
 ## Prose rules
 
--- a/README.md
+++ b/README.md
@@ -51,6 +51,101 @@ Install the full auto-instrumentation suite:
 sudo dnf install opentelemetry
 ```
 
+## Configuring where telemetry goes
+
+By default, every auto-instrumentation package exports OTLP to `localhost`.
+Point it at a real destination with one of the two options below.
+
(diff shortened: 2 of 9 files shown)


mmanciop:
Closing to reopen from a fork to avoid CLA issues with the branch.

PR #66, opened 2026-08-06
Drop elasticsearch instrumentation and bump Python bundle to 1.44.0/0.65b0

ocelotl:
## Summary

`opentelemetry-instrumentation-elasticsearch` was dropped upstream in
open-telemetry/opentelemetry-python-contrib#4759 because all supported
`elasticsearch` clients now ship native OpenTelemetry instrumentation. Its last
release is `0.64b0`, which hard-pins `opentelemetry-instrumentation==0.64b0`, so
keeping it in the bundle blocks every other package from moving to `0.65b0`
(this is what forced the temporary SDK revert in #64).

This PR:

- removes the `opentelemetry-instrumentation-elasticsearch` pin, and
- bumps the Python bundle to the current release now that the full set is
  resolvable:
  - `opentelemetry-sdk` `1.43.0` → `1.44.0`
  - `opentelemetry-distro`, `opentelemetry-instrumentation`,
    `opentelemetry-propagator-ot-trace` and all
    `opentelemetry-instrumentation-*` packages `0.64b0` → `0.65b0`

This reverses the temporary revert from #64.

## Validation

- Confirmed on PyPI that every bumped package has a `0.65b0` release
  (`opentelemetry-instrumentation-elasticsearch` is the only one that does not,
  hence its removal).
- `uv pip compile` resolves the full PyPI portion of the requirements cleanly
  (68 packages, no conflicts; API/SDK at `1.44.0`, contrib at `0.65b0`).

The recurring half of this problem — Renovate never advancing the
pre-release-only contrib pins — is addressed separately.

### DIFF
--- a/packaging/common/python/requirements.txt
+++ b/packaging/common/python/requirements.txt
@@ -4,12 +4,12 @@
 # with opentelemetry-exporter-otlp-proto-http replaced by opentelemetry-exporter-otlp-pyproto-http
 # (pure-Python protobuf, no google-protobuf C-extension dependency).
 
-opentelemetry-distro==0.64b0
+opentelemetry-distro==0.65b0
 
 # The file-configuration extra (pyyaml, jsonschema) enables declarative
 # configuration via OTEL_CONFIG_FILE; without it, setting that variable makes
 # the configurator crash with ModuleNotFoundError.
-opentelemetry-sdk[file-configuration]==1.43.0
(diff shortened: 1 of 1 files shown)


douglascamata (approved):

ocelotl:
Marking as draft to prevent accidental merging @mmanciop

x1unix (approved):

mmanciop (approved):

PR #69, opened 2026-08-07
Add a post-install sanity check (otel-instrumentation-check)

### LINKED ISSUE #61: Add a post-install sanity check
After installing the `opentelemetry` package and restarting an application, there is currently no quick way for a user to confirm that the injector is active and instrumentation is being applied — without first having a working OTLP receiver listening on `localhost:4317`/`4318`.

This gap shows up in documentation: the installation guide can only say "restart an application and confirm it emits telemetry", which requires a full observability backend to be in place before the user can verify the most basic thing — that the package installed correctly.

We should provide a lightweight verification step that works immediately after install, without requiring a backend. Maybe something based on config files and file exporters? It will likely be language-specific, so it relates with #60.

ocelotl:
Fixes #61.

## What

Adds `otel-instrumentation-check`, a lightweight command that answers two questions about auto-instrumentation on a host, **without a running OTLP receiver and without restarting an application**. It ships in the `opentelemetry-injector` package at `/usr/bin/otel-instrumentation-check`.

1. **Is this machine set up so newly started processes will be instrumented?** A static, machine-wide check that reads the same files a newly started process would consult.
2. **Of the processes running right now, which are already instrumented and which started before the package was installed and so must be restarted to pick it up?** A per-process runtime scan.

```console
$ otel-instrumentation-check
OpenTelemetry auto-instrumentation post-install check

[ ok ] injector library present: /usr/lib/opentelemetry/injector/libotelinject.so
[ ok ] injector active in /etc/ld.so.preload
[ ok ] Python agent registered (python.conf): /usr/lib/opentelemetry/python/glibc
[info] telemetry destination: http://localhost:4317 (gRPC) or http://localhost:4318 (HTTP), the default. No OTLP receiver is needed for this check, but one must be reachable for telemetry to arrive.
[info] running processes on supported runtimes: 1 monitored, 1 not monitored (restart required). A mapped injector proves the injector ran, not that the SDK is healthy; runtime detection is a best-effort name match.
[warn] NOT monitored: pid 3120 (python3, Python) started before install; restart it to instrument (e.g. systemctl restart the owning unit)
[ ok ] monitored: pid 4711 (java, Java)

Result: ACTION NEEDED. The injector is active, but some already-running processes started before install and must be restarted to be instrumented; see the [warn] items above.
```

### Static machine-wide check

Answering @atoulme's question on the issue ("Is a collector required for this? Can we check the collector status according to systemd?"): no collector is required, and there is deliberately no systemd check. The injector is an `LD_PRELOAD` library, not a daemon, so there is no service whose status could report "instrumentation is on". The check instead reads the exact files a newly started process would consult and reports what that process would pick up:

- **Injector active**: `/etc/ld.so.preload` lists `libotelinject.so` (matched as a whitespace-separated field, the same way the postinstall script and the dynamic linker treat the file), and the library exists on disk.
- **Language agents registered**: each `conf.d/*.conf` drop-in is parsed and the agent path it registers is checked for existence (prefix keys resolve the `glibc/` subdirectory, matching the injector).
- **Telemetry destination**: reports `OTEL_CONFIG_FILE` (declarative config) or the `OTEL_EXPORTER_OTLP_ENDPOINT` (falling back to the documented default), so a missing backend is never mistaken for a broken install.

### Per-process runtime scan

The static check confirms the install is correct, but it cannot tell you whether the services **running right now** are instrumented: `/etc/ld.so.preload` is consulted only at process startup, so a process that started before the package was installed is never instrumented until it restarts. This is the end-user question @mmanciop raised on the issue and the PR ("Which applications are now being monitored and which not, and have to be restarted to be monitored"), so the check now scans the running processes too.

For each running process on a supported runtime (Java, Node.js, .NET, Python), the scan reads `/proc/<pid>/maps` and tests whether `libotelinject.so` is mapped into it. That mapping is the ground-truth signal that the process went through the injector at startup:

- injector mapped -> the process started after install and is being instrumented;
- injector not mapped -> the process started before install and must be restarted (for example, `systemctl restart` the owning unit).

The scan is best-effort by design: it needs no restart, no OTLP receiver, and no injector logs, and it never fails the overall check because of a process it cannot read. Reading another user's `/proc/<pid>/maps` requires root, so when the scan cannot read a process it skips it and prints a hint to run as root for full coverage.

Two limitations are documented in the code, the report output, the README, and the man page:

- a mapped `libotelinject.so` proves the **injector ran** in the process, not that the language **SDK finished attaching and is healthy** (the injector could run and the agent still fail to load), so the running-process result is the best signal available without a running application, not a guarantee; and
- runtime detection is a simple substring match on the process command name and `argv[0]`, which can misfire on an unrelated wrapper name.

A definitive per-process answer would need a cross-language SDK health mechanism that does not exist in OpenTelemetry yet; this scan is the best available proxy today.

### Exit codes

- `0` the injector is active, at least one language agent is present, and no running process needs a restart;
- `1` an installation problem was found;
- `2` usage error;
- `3` the installation is correct but one or more already-running processes started before install and must be restarted to be instrumented.

Code `3` is distinct from `0` so a provisioning script can detect the "needs restart" case and restart the affected `systemd` units without treating it as an installation failure. An installation problem (`1`) outranks the needs-restart signal.

## Changes

- `cmd/otel-instrumentation-check/` - the command plus unit tests (temp-dir and fake-`/proc` based, mirroring `otel-config-check`): the static install check and the `/proc/<pid>/maps` running-process scan.
- Build wiring: cross-compiled like `otel-config-check` (`make otel-instrumentation-check`), embedded in the injector package by `packaging/builder`.
- `packaging/tests/lifecycle/sanitycheck_test.go` - container tests: a correct install exits `0` or `3` (both mean the install is wired up) and reports the running-process scan summary; the injector installed with no language agent exits `1`.
- Docs: main README, injector README and man page, CONTRIBUTING repository layout, the architecture design doc (filesystem layout, injector contents, file-ownership table), and the integration test plan - all updated to cover the running-process scan, exit code `3`, the run-as-root hint, and the documented limitations.

## Testing

- `make go-unit-tests` (adds the new package's suite, including the running-process scan tests)
- `go vet ./...` and `gofmt` clean
- Built the injector `.deb` locally and confirmed `/usr/bin/otel-instrumentation-check` is embedded (mode 0755)
- Container lifecycle tests added under `integration-test-{deb,rpm}-lifecycle`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,6 +16,7 @@ No Ruby, FPM, or special Docker images are required to build packages.
 cmd/build-packages/          CLI entry point for building .deb and .rpm packages
 cmd/build-spec/rpm/          CLI entry point for generating and staging the source RPM spec
 cmd/otel-config-check/       Declarative-config validator shipped inside the Python package
+cmd/otel-instrumentation-check/  Post-install sanity check shipped inside the injector package
 packaging/
   builder/                   Go library that drives nfpm to create packages
     builder.go               Build orchestration, common metadata
--- a/Makefile
+++ b/Makefile
@@ -71,24 +71,35 @@ TEST_LANGUAGES := java nodejs dotnet python
 otel-config-check:
 	CGO_ENABLED=0 GOOS=linux GOARCH=$(ARCH) go build -trimpath -o build/bin/otel-config-check-$(ARCH) ./cmd/otel-config-check
 
+# Cross-compiles the otel-instrumentation-check post-install sanity check that
+# ships inside the injector package. Same pure-Go, CGO-disabled build as
+# otel-config-check.
+.PHONY: otel-instrumentation-check
+otel-instrumentation-check:
+	CGO_ENABLED=0 GOOS=linux GOARCH=$(ARCH) go build -trimpath -o build/bin/otel-instrumentation-check-$(ARCH) ./cmd/otel-instrumentation-check
+
+# Both cross-compiled Go binaries the package builds embed.
+.PHONY: check-binaries
+check-binaries: otel-config-check otel-instrumentation-check
+
 .PHONY: deb-package-%
-deb-package-%: otel-config-check
+deb-package-%: check-binaries
 	go run ./cmd/build-packages -version $(VERSION) -arch $(ARCH) -format deb -component $* -output $(OUTPUT_DIR)
 
 .PHONY: deb-packages
-deb-packages: otel-config-check
+deb-packages: check-binaries
 	go run ./cmd/build-packages -version $(VERSION) -arch $(ARCH) -format deb -output $(OUTPUT_DIR)
 
 .PHONY: rpm-package-%
-rpm-package-%: otel-config-check
+rpm-package-%: check-binaries
 	go run ./cmd/build-packages -version $(VERSION) -arch $(ARCH) -format rpm -component $* -output $(OUTPUT_DIR)
 
 .PHONY: rpm-packages
-rpm-packages: otel-config-check
+rpm-packages: check-binaries
 	go run ./cmd/build-packages -version $(VERSION) -arch $(ARCH) -format rpm -output $(OUTPUT_DIR)
 
 .PHONY: packages
-packages: otel-config-check
+packages: check-binaries
 	go run ./cmd/build-packages -version $(VERSION) -arch $(ARCH) -format all -output $(OUTPUT_DIR)
 
 # ============================================================================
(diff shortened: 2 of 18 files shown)


mmanciop:
@ocelotl you where faster with the PR and my writing down my thoughts on the matter. See https://github.com/open-telemetry/opentelemetry-packaging/issues/61#issuecomment-5224783421 .

I think the scope here is different than what I had in mind in #61: I meant it as "Which applications are now being monitored and which not (and have to be restarted to be monitored)", which is what end users ultimately need to know so they can, e.g., restart systemd units of processes that were created before the package was installed.

Issue #52, opened 2026-07-29
Add jmx-scraper package


atoulme:
This is the follow-up of https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1964 now that this SIG exists.

The Java SIG maintainers has stated in issues and SIG meetings that they don't want to deal with the packaging aspect of the software, and would rather have the packaging SIG take over.

I am offering that we start to curate a new package for the jmx scraper that easily allows us to wrap this jar file into a deliverable for a Linux host to start with, as a Debian or RPM package, and learn from this limited experience.

PR #70, opened 2026-08-09
This adds to the spec a new component, opentelemetry-jmx-scraper, for inclusion

atoulme:
See #52 for context.

This PR is just the specification changes related to this new package.

### DIFF
--- a/docs/design/packages-meta-architecture.md
+++ b/docs/design/packages-meta-architecture.md
@@ -28,7 +28,8 @@ This document describes all packages in the first version of the system packages
 
 ## Packages overview
 
-The first version ships six packages, each available as both DEB and RPM.
+The first version ships seven packages, each available as both DEB and RPM.
+Six of them form an integrated suite around the injector; the seventh, `opentelemetry-jmx-scraper`, is fully independent — see [Independent packages](#independent-packages).
 
 | Package | Description | Architecture |
 |---------|-------------|-------------|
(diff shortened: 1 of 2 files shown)


docs/design/packages-meta-architecture.md
```diff
@@ -324,6 +341,78 @@ If `Depends` were used instead, removing any single language package would force
 
 See [Injector interface versioning](#injector-interface-versioning) for the upgrade scenario.
 
+### `opentelemetry-jmx-scraper`
```
mmanciop, line 344: We should mention why we do not have relations between this package and `opentelemetry-java-autoinstrumentation` (you do not need `opentelemetry-jmx-scraper` to monitor JVMs from the outside if you have the OpenTelemetry Java Agent inside).

docs/design/packages-meta-architecture.md
```diff
@@ -324,6 +341,78 @@ If `Depends` were used instead, removing any single language package would force
...
+
+```ini
+EnvironmentFile=-/etc/opentelemetry/jmx-scraper/jmx-scraper.env
+ExecStart=${JAVA_HOME}/bin/java -jar /usr/lib/opentelemetry-jmx-scraper/opentelemetry-jmx-scraper.jar -config ${JMX_SCRAPER_CONFIG}
+```
+
+`jmx-scraper.env` ships with both variables set to their defaults:
+
+```sh
+JAVA_HOME=/usr/lib/jvm/default-java
```
mmanciop, line 387: This is Debian-specific. On RHEL, this would result on the service failing. Needs to be handled somehow, e.g. fall back to `$(readlink -f $(which java))` or (perhaps better) `ExecStart=/usr/bin/env java -jar ...` in the Systemd Unit.

docs/design/packages-meta-architecture.md
```diff
@@ -324,6 +341,78 @@ If `Depends` were used instead, removing any single language package would force
...
+JMX_SCRAPER_CONFIG=/etc/opentelemetry/jmx-scraper/config.properties
```
mmanciop, line 388: By default, the scraper does not do anything useful, needs to be pointed at the right process on the right port (+ auth if needed). Yet, the Systemd Unit will start automatically, giving a false sense of security and much FUD. This is something that needs to be handled IMO in two says:
(1) Some sort of auto-discovery looking at java processes with `-Dcom.sun.management.jmxremote` for headless installs
(2) A configuration step _during the installation_ for non `-y` installations; pretty much like the Java packages ask you questions about timezone and stuff.

mmanciop (changes requested):

PR #76, opened 2026-08-11
Verify .NET agent checksum against upstream checksums.txt

mwimpelberg28:
The build downloaded the glibc archive from GitHub releases with no integrity check beyond TLS. Every release publishes a checksums.txt alongside the archives, so fetch it and verify the downloaded zip's SHA-256 before extraction.


mmanciop (approved):

mmanciop:
Thanks @mwimpelberg28 !

Issue #21, opened 2026-07-10
Verify checksums/signatures of downloaded upstream artifacts at build time


mmanciop:
The package build downloads upstream artifacts over HTTPS and bundles them without verifying integrity beyond TLS: the injector .so, Java agent JAR, and .NET ZIPs from GitHub releases, Node.js via npm, Python via pip (all in packaging/builder/download.go). Links back to the review thread on #18.

We should pin expected checksums alongside the existing release.txt pins so Renovate updates both together; verify against upstream checksum/signature assets where published; use ecosystem-native integrity for npm/pip (lockfile integrity, pip --require-hashes); and consider embedding the artifact checksums in the packages themselves so bundled provenance is auditable after installation.

mwimpelberg28:
@mmanciop I put in this PRs
Java https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/19914

PR #77, opened 2026-08-12
Verify Java agent GPG signature against upstream release key

mwimpelberg28:
## Summary

- `downloadJavaAgent` downloaded `opentelemetry-javaagent.jar` from GitHub releases with no integrity check beyond TLS, the same gap #76 fixed for the .NET agent.
- Unlike .NET, the Java release doesn't publish a `checksums.txt` — it publishes a detached GPG signature (`opentelemetry-javaagent.jar.asc`) instead. Verify the jar against it using the OpenTelemetry Java release key (fingerprint `3F05DDA9F317301E927136D417A27CE7A60FF5F0`), pinned in-repo at `packaging/common/java/release-key.asc`.
- The fingerprint isn't published through any OpenTelemetry-controlled channel (no `KEYS` file, no docs page); it's only discoverable on public keyservers because Maven Central's Sonatype Central Portal requires keyserver presence for signed uploads. Trust here is necessarily trust-on-first-use — this fingerprint was confirmed to have signed the jar consistently across releases spanning roughly two years (v2.10.0, v2.20.1, v2.30.0) before being pinned. Documented in the code comment above `javaReleaseKeyFingerprint`.
- Also wires `packaging/builder`'s tests (including the existing .NET checksum tests from #76) into `make go-unit-tests`, which previously only ran `./cmd/...` and never executed them in CI.

Related: #21

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make go-unit-tests` (now covers `packaging/builder`, including the new signature-verification tests)
- [x] Verified end-to-end against the real v2.30.0 release: genuine jar+signature pass; a byte-tampered copy of the jar is correctly rejected


mmanciop (changes requested):
This change is brittle against the Java SIG changing signing key or method. I think it'd be best to involve the Java SIG to publish a checksum file, and then consume it here.

PR #81, opened 2026-08-20
Track the minimum supported Python and keep the sitecustomize gate in sync

### LINKED ISSUE #78: Track which version of Python is supported by the automatic instrumentation and have automation to update the sitecustomize
We currently hard-code the minimum Python version in sitecustomize: https://github.com/open-telemetry/opentelemetry-packaging/blob/0e45b75cfac1d269f8aabc61d2d6d15435ff8551/packaging/common/python/sitecustomize.py#L242

We should have a mechanism to update that when we upgrade the instrumentation to a version that raises the minimum supported version.

ocelotl:
## What

Derive the minimum supported Python of the bundled agent from the strictest `Requires-Python` across the bundled distributions, and keep the `sitecustomize.py` version gate in sync with it instead of relying on a hand-maintained magic number.

Closes #78.

## Why

The floor was hard-coded in `sitecustomize.py` (`version_info[1] < 10`) with nothing tying it to the distributions we actually ship. When a dependency bump raises the minimum supported Python, that gate silently drifts.

## How

- Refactor the gate into a named constant with an anchor comment, so it stays a plain integer comparison that still parses and runs on ancient interpreters, while giving the tooling an unambiguous target:

  ```python
  _MINIMUM_PYTHON_MINOR = 10  # sync-minimum-python-version: 3.x minor floor
  if version_info[0] != 3 or version_info[1] < _MINIMUM_PYTHON_MINOR:
  ```

- Add `packaging/common/python/sync_minimum_python_version.py`. It derives the floor as the strictest `Requires-Python` lower bound over every installed distribution plus every vendored `pyproject.toml`, then either verifies the gate (`--check`) or rewrites it (`--write`).
- Add `make check-minimum-python-version`, wired into the `unit-tests` CI job. It installs the PyPI pins into a throwaway venv purely to read their metadata, and reads the vendored floor straight from each vendor `pyproject.toml`.
- Add offline unit tests for the derivation, gate parsing, and rewrite logic (discovered by `make python-unit-tests`).

## Design note (feedback welcome)

The check venv is built with the minimum supported interpreter (3.10) on purpose, so pip resolves transitive dependencies the way they resolve on the floor. A newer `rpds-py` release requiring 3.11 otherwise inflated the derived floor under a modern host.

The effect: this is robust detection plus a convenience `--write` for the current floor, not fully automatic raising. When a pinned package actually raises its `Requires-Python` above the floor, `pip install` under 3.10 fails and CI goes red, and a maintainer bumps the gate and `MINIMUM_PYTHON` deliberately. That keeps dropping a supported Python version a conscious decision rather than a silent bot edit. Happy to switch to direct-pin-only derivation under a modern interpreter if maintainers prefer `--write` to auto-raise.

## Testing

- `make python-unit-tests` (new tests discovered and passing).
- `make check-minimum-python-version` derives 3.10, matches the gate, exits 0.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,6 +43,14 @@ jobs:
       - name: Run the vendored pyproto exporter test suites
         run: make pyproto-unit-tests
 
+      - name: Set up the minimum supported Python for the version-gate check
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: "3.10"
+
+      - name: Check the sitecustomize minimum Python version is in sync with the bundled distributions
+        run: make check-minimum-python-version
+
   sitecustomize-compatibility-tests:
     runs-on: ubuntu-24.04
     steps:
--- a/Makefile
+++ b/Makefile
@@ -471,6 +471,31 @@ pyproto-unit-tests:
 			$(PYPROTO_VENDOR_DIR)/$$suite; \
 	done
 
+# Check that the sitecustomize.py version gate stays in sync with the strictest
+# Requires-Python across the bundled distributions. The PyPI pins are installed
+# into a throwaway venv (excluding the vendored source lines) purely so the tool
+# can read their Requires-Python metadata; the vendored floor is read straight
+# from each vendor pyproject.toml via --vendor-dir, so the vendored source does
+# not need to be built for the check.
+#
+# The venv is built with the minimum supported interpreter (MINIMUM_PYTHON, the
+# current 3.x floor) on purpose: pip must resolve transitive dependencies the
+# way it would on that floor, so a newer release of a transitive dependency that
+# raised its own Requires-Python does not inflate the derived floor above what
+# actually runs on the minimum interpreter. tomli is the tomllib backport the
+# tool falls back to under Python 3.10 (tomllib is standard library from 3.11).
+MINIMUM_PYTHON ?= python3.10
+.PHONY: check-minimum-python-version
+check-minimum-python-version:
+	$(MINIMUM_PYTHON) -m venv build/minimum-python-version-venv
+	build/minimum-python-version-venv/bin/pip install --quiet packaging tomli
+	grep -v '^\./vendor/' packaging/common/python/requirements.txt | \
+		build/minimum-python-version-venv/bin/pip install --quiet -r /dev/stdin
+	build/minimum-python-version-venv/bin/python \
+		packaging/common/python/sync_minimum_python_version.py --check \
+		--vendor-dir packaging/common/python/vendor \
+		--sitecustomize packaging/common/python/sitecustomize.py
+
 # ============================================================================
 # Lint
 # ============================================================================
--- a/Makefile
+++ b/Makefile
@@ -400,13 +400,15 @@ integration-test-rpm-vendor: local-rpm-repo local-rpm-vendor-repo
 go-unit-tests:
 	go test -v ./cmd/...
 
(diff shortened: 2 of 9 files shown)


mmanciop (changes requested):
In my tests, `distributions()` enumerates everything in the venv, including packages the DEB and RPM never ship. I created a deliberately old-pin venv, where the derived floor was 3.7. I then installed setuptools==84.0.0 (Requires-Python >=3.10), changed nothing else, and the derivation jumped to 3.10:

derived minimum Python: 3.10
sitecustomize gate: 3.7

The Makefile venv always contains four such packages: pip (>=3.9), setuptools (>=3.9), and the tool's own packaging (>=3.9) and tomli (>=3.8). Today they all sit below 3.10, so the check is right by luck. The failure mode is one-directional and silent: scaffolding can only push the floor up, so it masks a floor that should drop, and --check reports "in sync" while the gate excludes interpreters the bundle actually supports. Since the venv runs on MINIMUM_PYTHON, scaffolding tends to require exactly that minor, which is the worst case for masking.

I would filter the enumeration to the distributions named in requirements.txt rather than everything importable. That also removes the reliance on a pristine venv.

ocelotl:
> In my tests, `distributions()` enumerates everything in the venv, including packages the DEB and RPM never ship. I created a deliberately old-pin venv, where the derived floor was 3.7. I then installed setuptools==84.0.0 (Requires-Python >=3.10), changed nothing else, and the derivation jumped to 3.10:
> 
> derived minimum Python: 3.10 sitecustomize gate: 3.7
> 
> The Makefile venv always contains four such packages: pip (>=3.9), setuptools (>=3.9), and the tool's own packaging (>=3.9) and tomli (>=3.8). Today they all sit below 3.10, so the check is right by luck. The failure mode is one-directional and silent: scaffolding can only push the floor up, so it masks a floor that should drop, and --check reports "in sync" while the gate excludes interpreters the bundle actually supports. Since the venv runs on MINIMUM_PYTHON, scaffolding tends to require exactly that minor, which is the worst case for masking.
> 
> I would filter the enumeration to the distributions named in requirements.txt rather than everything importable. That also removes the reliance on a pristine venv.

Right, fixed that in the latest commit :v:

.github/workflows/build.yml
```diff
@@ -43,6 +43,14 @@ jobs:
       - name: Run the vendored pyproto exporter test suites
         run: make pyproto-unit-tests
 
+      - name: Set up the minimum supported Python for the version-gate check
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: "3.10"
```
mmanciop, line 49: Does this need to match the minimum version?

Makefile
```diff
@@ -471,6 +473,43 @@ pyproto-unit-tests:
...
+# payload's transitive dependencies the way it would on that floor, so a newer
+# release of a transitive dependency that raised its own Requires-Python does
+# not inflate the derived floor above what actually runs on the minimum
+# interpreter. tomli is the tomllib backport the tool falls back to under
+# Python 3.10 (tomllib is standard library from 3.11).
+MINIMUM_PYTHON ?= python3.10
+MINIMUM_PYTHON_VENV = build/minimum-python-version-venv
+MINIMUM_PYTHON_PAYLOAD = build/minimum-python-version-payload
+.PHONY: check-minimum-python-version
+check-minimum-python-version:
```
mmanciop, line 483: Please extract this to a script in .github/scripts/

ocelotl: fixed :v:

PR #54, opened 2026-07-30
Fedora COPR support PoC

x1unix:
Add SRPM spec generator and build scripts for COPR

COPR repo: https://copr.fedorainfracloud.org/coprs/x1unix/opentelemetry-packaging/
Build: https://github.com/x1unix/opentelemetry-packaging/actions/runs/30514516950

The repo mirroring part is done in separate branch: https://github.com/x1unix/opentelemetry-packaging/tree/cloudflare-r2-mirror

Related to: #4

PR #85, opened 2026-08-25
feat: Add SRPM spec generation support

x1unix:
Adds [SRPM spec](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/rpm_packaging_guide/packaging-software#what-is-a-spec-file_rpm-packages) generation and local COPR-style rebuild support.

This PR is a prerequisite for Fedora COPR build support.
Related to: https://github.com/open-telemetry/opentelemetry-packaging/issues/4

- Generates an RPM spec and vendored source archive from shared package metadata.
- Adds native and containerized SRPM build targets.
- Adds Fedora and EL rebuild validation plus metadata tests.
- Preserves existing nfpm DEB/RPM builds.


cmd/build-packages/main.go
```diff
@@ -33,20 +40,35 @@ func main() {
...
+	format := flag.String("format", "all", "Package format: deb, rpm, spec, or all")
 	outputDir := flag.String("output", "build/packages", "Output directory for built packages")
-	component := flag.String("component", "all", "Component to build: injector, java, nodejs, dotnet, meta, or all")
+	component := flag.String("component", "all", "Component to build: injector, java, nodejs, dotnet, python, meta, or all")
 	packagingDir := flag.String("packaging-dir", "", "Path to packaging/ directory (auto-detected if empty)")
 	configCheckBinary := flag.String("config-check-binary", "",
 		"Path to a prebuilt otel-config-check binary for the target architecture "+
 			"(defaults to build/bin/otel-config-check-<arch>; build it with `make otel-config-check`)")
+	stageRoot := flag.String("stage-root", "",
+		"Stage the payload into this directory (an rpmbuild %buildroot) instead of building a package")
```
mmanciop: Is this something we could also use in deb? If so, could you please make the message more generic and integrate the option in the deb build as well?

cmd/build-packages/main.go
```diff
@@ -33,20 +40,35 @@ func main() {
...
+	filelistDir := flag.String("filelist-dir", "",
+		"Directory for the generated RPM %files fragments; required with -stage-root")
+	printRPMVersion := flag.Bool("print-rpm-version", false,
```
mmanciop: I think we should avoid hard-coding so much RPM specifics. This could be `print-packager-version` and report RPM or DEB as needed

cmd/build-packages/main.go
```diff
@@ -114,6 +126,55 @@ func main() {
...
+			}
+		}
+		fmt.Println("All components staged successfully")
+		return
+	}
+
+	var formats []string
+	switch *format {
+	case "all":
+		formats = []string{"deb", "rpm"}
```
mmanciop, line 101: No `spec` here? That is not longer `all` then, but `packages` maybe?

cmd/build-packages/main.go
```diff
@@ -114,6 +126,55 @@ func main() {
...
+	case "deb", "rpm":
+		formats = []string{*format}
+	case "spec":
```
mmanciop: I have a hard time understand what `spec` means, I don't think it is self-descriptive enough

packaging/builder/builder.go
```diff
@@ -38,12 +43,83 @@ type Config struct {
...
-	// a cleanup function that removes any staging directories, and an error.
-	// The cleanup function must be called after packaging completes.
-	InfoFunc func(cfg Config, format string) (info *nfpm.Info, cleanup func(), err error)
+	// Noarch marks a package whose payload is architecture independent. It
+	// selects the "all" DEB architecture and the "noarch" RPM one.
+	Noarch bool
+	// Provides, Depends, Recommends, and Suggests declare the package
+	// relationships. They all use the interface-versioned virtual names
+	// described in docs/design/packages-meta-architecture.md.
+	Provides   []string
```
mmanciop: Maybe we nest this in a `Relations` substruct?

mmanciop (changes requested):
I overall like the PR a bunch. I think, however, it would be more future-proof with the adjustments I commented about

cmd/build-packages/main.go
x1unix: @mmanciop this is a RPM-specific flag that will be used in the next PR.

Context: https://github.com/open-telemetry/opentelemetry-packaging/pull/54/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R124 

It will be needed by copr-related makefile targets to build SRPM.

I've exposed this as a flag to avoid duplicating the same version normalization logic in Go and bash.

cmd/build-packages/main.go
x1unix: @mmanciop I didn't add SRPM spec to `all` as it doesn't produce a package, but a recipe to build one (like `dpkg-buildpackage --build=source` does).

I will probably move SRPM spec generator into a separate binary to avoid confusion here.

If we adopt Launchpad, we would need to make source packages for debian/ubuntu as well and it also might have deb-specific logic and flags.

cmd/build-packages/main.go
x1unix: Update: I've decoupled srpm spec generator into a separate tool to avoid confusion: https://github.com/open-telemetry/opentelemetry-packaging/pull/85/changes/0d4a1ae78e478c6d9c7ea67004e8ccc739236174

cmd/build-packages/main.go
x1unix: https://github.com/open-telemetry/opentelemetry-packaging/pull/85/changes/0d4a1ae78e478c6d9c7ea67004e8ccc739236174

packaging/builder/builder.go
x1unix: @mmanciop done: https://github.com/open-telemetry/opentelemetry-packaging/pull/85/changes/ee8e4d473ae32c3262334158e5ab31ef20abdd6e

cmd/build-packages/main.go
x1unix: @mmanciop if deb source packages generation (launchpad) will be added, I guess some flags can be reused for both package spec formats, but I'm not sure as I don't have a full picture yet.

x1unix:
@mmanciop thanks for a review.

I've reverted the changes to `cmd/build-packages/` and moved those into a dedicated `cmd/build-spec` tool as it contains a lot of SRPM-specific logic, that don't belong to a DEB/RPM package builder.

cmd/build-packages/main.go
```diff
@@ -40,35 +33,20 @@ func main() {
 
 	version := flag.String("version", "", "Package version (required)")
 	arch := flag.String("arch", "amd64", "Target architecture (amd64 or arm64)")
-	format := flag.String("format", "all", "Package format: deb, rpm, spec, or all")
+	format := flag.String("format", "all", "Package format: deb, rpm, or all")
 	outputDir := flag.String("output", "build/packages", "Output directory for built packages")
-	component := flag.String("component", "all", "Component to build: injector, java, nodejs, dotnet, python, meta, or all")
+	component := flag.String("component", "all", "Component to build: injector, java, nodejs, dotnet, meta, or all")
```
mmanciop: Python here was right, no?

cmd/build-spec/rpm/main.go
```diff
@@ -0,0 +1,164 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Command build-spec drives the spec-based source RPM build used by COPR.
```
mmanciop, line 4: I’d change the path to cmd/build-spec/rpm until we know if it is portable to other build systems

mmanciop (approved):

cmd/build-spec/rpm/main.go
x1unix: Done in https://github.com/open-telemetry/opentelemetry-packaging/pull/85/changes/eaadfeded72ae41ecda818887168e15459b8e55f

cmd/build-packages/main.go
x1unix: @mmanciop done

PR #88, opened 2026-08-30
Verify Python package checksums against PyPI's published digests

mwimpelberg28:
## Summary

`downloadPythonAgent` installs wheels/sdists straight from PyPI with no integrity check beyond TLS. PyPI publishes a sha256 digest for every release file via its JSON API, so this downloads the resolved dependency closure (pinned requirements and their transitive dependencies alike) into a local cache first, verifies every file against PyPI's own published digest, and only then installs `--no-index` from that verified cache — the same trust model `downloadDotnetAgent` (#76) uses for the .NET `checksums.txt`: the expected digest is fetched fresh from the upstream registry for the resolved version rather than stored in this repo, so Renovate only needs to keep bumping `requirements.txt`.

Unlike the Java agent JAR / .NET zips (raw GitHub release binaries with no built-in checksums), and unlike a static `checksums.txt` published by opentelemetry-python's own release workflow, PyPI already publishes a checksum for every file it hosts — so this covers the whole resolved bundle, including transitive dependencies that aren't explicitly pinned in `requirements.txt`, with no changes needed upstream.

Related to #21.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` clean
- [x] `go test ./packaging/builder/...` (new unit tests for filename parsing, PyPI digest lookup, and cache verification, plus existing tests)
- [ ] Docker-based integration tests (`packaging/tests/python`) — not runnable in this sandbox (no Docker access); exercised by CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -119,6 +119,17 @@ jobs:
           path: ./build/packages
 
       - name: Run packaging integration tests for ${{ matrix.SYS_PACKAGE }}/${{ matrix.lang }}/${{ matrix.arch }}
+        env:
+          # local-apt-repo (a prerequisite of every integration-test-*
+          # target) rebuilds the packages locally via the .PHONY
+          # deb-packages/rpm-packages targets, re-running
+          # downloadInjector/downloadJavaAgent's GitHub asset-digest lookups
+          # even though a prebuilt package was already downloaded above.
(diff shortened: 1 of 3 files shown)


mmanciop (approved):

mmanciop:
@mwimpelberg28 LGTM and thanks for doing this. It needs a few conflicts to be resolved before merging.

PR #90, opened 2026-08-31
Fix: thread GITHUB_TOKEN into integration-test job too

mwimpelberg28:
## Summary

Follow-up to #89, which added GitHub asset-digest lookups (`fetchGitHubAssetDigest`, hitting `api.github.com`) to `downloadInjector`/`downloadJavaAgent`, and threaded `GITHUB_TOKEN` into the `build-packages` job so those lookups run authenticated.

It missed that `local-apt-repo` — a prerequisite of every `integration-test-*` Makefile target — depends on the `.PHONY` `deb-packages`/`rpm-packages` targets, so **every** integration-test job rebuilds the packages locally too, regardless of which language is under test, re-running the same digest lookups even though a prebuilt package was already downloaded in the prior step. Without a token there, those lookups run unauthenticated and share the 60/hour GitHub API rate limit across the runner IP pool — easy to exhaust once enough of the 24-job integration matrix (2 SYS_PACKAGE × 6 lang × 2 arch) runs concurrently:

```
error building deb injector: building info for injector: downloading injector:
fetching asset digest: fetching release metadata for
open-telemetry/opentelemetry-injector@v0.11.0: HTTP 403
```

Observed failing on #88's CI (`integration-tests (deb, python, amd64)`, `integration-tests (deb, lifecycle, amd64)`), but the bug lives in the already-merged `build.yml` from #89, so it currently affects every PR's `integration-tests` job, not just #88's.

This threads `GITHUB_TOKEN` into that job's step the same way `build-packages` already gets it.

## Test plan
- [x] YAML syntax validated
- [ ] CI on this PR itself exercises the fix (`integration-tests` jobs should stop hitting HTTP 403)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,15 +120,6 @@ jobs:
 
       - name: Run packaging integration tests for ${{ matrix.SYS_PACKAGE }}/${{ matrix.lang }}/${{ matrix.arch }}
         env:
-          # local-apt-repo (a prerequisite of every integration-test-*
-          # target) rebuilds the packages locally via the .PHONY
-          # deb-packages/rpm-packages targets, re-running
-          # downloadInjector/downloadJavaAgent's GitHub asset-digest lookups
-          # even though a prebuilt package was already downloaded above.
-          # Without a token those lookups share the 60/hour unauthenticated
-          # GitHub API rate limit across every concurrent job in this
-          # matrix (and the rest of the runner IP pool), which is easy to
-          # exhaust; see the build-packages job for the same reasoning.
           GITHUB_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -119,6 +119,17 @@ jobs:
           path: ./build/packages
 
       - name: Run packaging integration tests for ${{ matrix.SYS_PACKAGE }}/${{ matrix.lang }}/${{ matrix.arch }}
+        env:
+          # local-apt-repo (a prerequisite of every integration-test-*
```
mmanciop: @mwimpelberg28 could you please remove this not particularly-useful comment?

mmanciop (changes requested):

.github/workflows/build.yml
mwimpelberg28: Removed — thanks!

mmanciop (approved):

Issue #96, opened 2026-09-04
sitecustomize.py: diagnostics carry a severity, so they should carry a logging level instead of a print prefix


ocelotl:
`sitecustomize.py` writes its diagnostics with `print`, but it is already using the concept of severity. That concept has a first-class representation in the standard library, and we should use it.

Today the severity is a string that gets interpolated into a `print` call:

https://github.com/open-telemetry/opentelemetry-packaging/blob/f569839889abe2f1cee92fb6a12794ccfa7e3116/packaging/common/python/sitecustomize.py#L47-L65

Look at what that code is made of:

| what it does | where | what implements the same thing |
|---|---|---|
| names a severity (`WARN`, `DEBUG`) | lines 60, 64 | `logging` levels |
| filters by severity | line 44, line 64 | `Logger.setLevel` |
| formats a record with a fixed prefix | line 54 | `logging.Formatter` |
| writes to a stream, tolerating a `None` or broken one | lines 51-56 | `logging.StreamHandler` and `Handler.handleError` |

That is a logger, hand-rolled. The severity is the load-bearing part: it is what the operator reads to decide whether the line is a problem or a trace, and it is what `OTEL_INJECTOR_LOG_LEVEL` claims to select on. Expressing it as a substring of a formatted line means the concept exists only by convention. `logging` represents it as an ordered, comparable level on a record, which is what we actually mean.

Two of the consequences are already visible:

- `OTEL_INJECTOR_LOG_LEVEL` is a boolean wearing a level's name (line 44). The comparison is exact, so `DEBUG` and `Debug` do nothing, and no value silences the `WARN` lines. With real levels, honoring `info`, `warning`, and `error` becomes a two-line change.
- `stderr` is bound at import (line 18) and the write is wrapped in a bare `except Exception: pass` (lines 55-56). Both are things a `StreamHandler` already does, and does better.

## Why the textbook library shape does not work here

The usual advice for a library is to call `logging.getLogger(__name__)`, install no handler, and let the application decide. That advice fails for this file, and the failure is silent, so it is worth writing down.

`logging` applies two independent filters, both sitting at `WARNING` by default:

1. **The effective level.** A new logger is created at `NOTSET` and inherits from root, whose default level is `WARNING`. `log.debug(...)` returns inside `isEnabledFor` before any handler is consulted.
[...]

Issue #93, opened 2026-09-04
sitecustomize.py: a failure inside auto_instrumentation.initialize() never deactivates the agent


ocelotl:
`sitecustomize.py` calls `auto_instrumentation.initialize()` with no arguments:

https://github.com/open-telemetry/opentelemetry-packaging/blob/f569839889abe2f1cee92fb6a12794ccfa7e3116/packaging/common/python/sitecustomize.py#L323-L331

Since `opentelemetry-instrumentation` v0.55b0 the signature is `initialize(*, swallow_exceptions: bool = True)`, and `_initialize` wraps the four load steps in a broad handler that logs the exception and returns normally unless the caller opts out:

```python
    try:
        distro = _load_distro()
        distro.configure()
        _load_configurators()
        _load_instrumentors(distro)
    except Exception as exc:  # pylint: disable=broad-except
        _logger.exception("Failed to auto initialize OpenTelemetry")
        if not swallow_exceptions:
            raise exc
```

We pass no arguments, so we get the default and the exception never leaves contrib. That makes the `except Exception` on line 327 unreachable for anything raised by `_load_distro`, `distro.configure`, `_load_configurators`, or `_load_instrumentors`. It still covers the `import` on line 325, which is why the handler is not obviously dead.

The result is that a whole class of initialization failure produces no deactivation and no warning from us.

## How this is reached

The clearest trigger is a declarative configuration file that `otel-config-check` accepts and the SDK then rejects. That is not a defect in the validator: it deliberately checks only that the file is readable, parses as YAML, is not empty, and declares `file_format: "1.0"`, per its own comment.

https://github.com/open-telemetry/opentelemetry-packaging/blob/f569839889abe2f1cee92fb6a12794ccfa7e3116/cmd/otel-config-check/main.go#L54-L57

The SDK validates the entire document against the `OpenTelemetryConfiguration` JSON schema, so any file that is well formed but wrong in its body passes our check and fails theirs.

## Reproduction

Package versions as pinned: `opentelemetry-instrumentation` 0.65b0, `opentelemetry-sdk` 1.44.0, `opentelemetry-configuration` 0.65b0, `jsonschema` 4.26.0. Bundle-shaped layout, with `sitecustomize.py` in `<prefix>/glibc` and `otel-config-check` at `<prefix>/`, both on the paths the code expects.

`OTEL_CONFIG_FILE` points at:

```yaml
file_format: "1.0"
tracer_provider: 42
```

[...]

PR #99, opened 2026-09-04
fix(python): deactivate when an unexpected error escapes import_distro()

### LINKED ISSUE #98: sitecustomize.py: an unexpected error anywhere in import_distro() leaves the agent active and retried by every child process
`sitecustomize.py` guards individual operations carefully, and then calls `import_distro()` with nothing around it:

https://github.com/open-telemetry/opentelemetry-packaging/blob/f569839889abe2f1cee92fb6a12794ccfa7e3116/packaging/common/python/sitecustomize.py#L337

An exception from anywhere in that call that is not one of the four narrow `try` blocks escapes into `site.execsitecustomize()`, and the process ends up in the exact state every guard in the file exists to prevent. This is the same failure class as #93, reached through paths #94 does not cover. #93 closed one door into that state; this is the door frame.

## Reproduction

The most reachable trigger is `dist._path` on line 111. That attribute is private and only `PathDistribution` has it, so a distribution surfaced by any other finder raises `AttributeError`. Nothing catches it, and nothing would notice in review either: the only test that exercises the line assigns `_path` to a `MagicMock` (`test_sitecustomize.py:396`), which accepts any attribute.

I reproduced it with the packaged script, unmodified, on `python:3.13-slim`. A `.pth` file installed in `site-packages` replaces `importlib.metadata.distributions` with one returning a `Distribution` that has `metadata["Name"] == "opentelemetry-sdk"` and no `_path`. The `.pth` is only a way to get in front of `sitecustomize.py`, since `site` processes `site-packages/*.pth` before `execsitecustomize()`.

```
Error in sitecustomize; set PYTHONVERBOSE for traceback:
AttributeError: '_DistributionWithoutPath' object has no attribute '_path'
Error in sitecustomize; set PYTHONVERBOSE for traceback:
AttributeError: '_DistributionWithoutPath' object has no attribute '_path'
  child ran to completion
--- state left behind in the parent ---
  site on sys.path:                                False
  site in os.environ['PYTHONPATH']:                True
  PYTHON_AUTO_INSTRUMENTATION_AGENT_PATH_PREFIX:   '/usr/lib/opentelemetry/python'
  opentelemetry modules imported:                  0
APP RAN TO COMPLETION
exit code: 0
```

| | expected on a deactivation | actual |
|---|---|---|
| site in `PYTHONPATH` | removed | still present |
| `PYTHON_AUTO_INSTRUMENTATION_AGENT_PATH_PREFIX` | cleared | still set |
| site on `sys.path` | removed | removed, but see below |
| `[opentelemetry-python-autoinstrumentation]` lines | 1 | 0 |
| child process repeats the failure | no | yes |
| application exit code | 0 | 0 |

## Why it matters

1. **No message that names the agent.** `site.execsitecustomize()` catches the exception and prints one line with no `[opentelemetry-python-autoinstrumentation]` prefix and no traceback unless `PYTHONVERBOSE` is set. Anything grepping for the agent's marker, including this repository's own container tests, sees silence.

2. **The deactivation never runs.** `PYTHONPATH` keeps the site and the injector prefix stays set, so every child process that execs a new interpreter repeats the failure and reprints the line. The second error line above is the child. On a gunicorn or celery deployment that multiplies by the worker count. This is exactly what `_self_deactivate` exists to prevent.

3. **`sys.path` is left inconsistent.** The failure lands between the removal on line 291 and the re-append on line 299, so the parent process has neither instrumentation nor the bundle on `sys.path`. #93's failure at least left the bundle importable. Here an application that legitimately imports something from the bundle after startup cannot.

4. **The application is unharmed and exits 0**, which is the only good news, and is what lets this go unnoticed indefinitely.

## Proposed fix

An outer guard whose own reporting is guarded, because the thing that failed can be the diagnostics channel:

```python
try:
    import_distro()
except Exception as unexpected_error:
    try:
        _self_deactivate(dirname(__file__))
        _print_cannot_auto_instrument_message(
            "unexpected error while deciding whether to auto-instrument: {}: {}".format(
                type(unexpected_error).__name__, unexpected_error))
    except Exception:
        pass
```

Same reproduction against the patched script:

```
[opentelemetry-python-autoinstrumentation] WARN: cannot auto-instrument Python process: unexpected error while
deciding whether to auto-instrument: AttributeError: '_DistributionWithoutPath' object has no attribute '_path' [/app.py]
  site in os.environ['PYTHONPATH']:                False
  PYTHON_AUTO_INSTRUMENTATION_AGENT_PATH_PREFIX:   ''
```

One prefixed line, a clean deactivation, no child retry. It also compiles on `python:2.7-slim`, so the parse contract on line 11 holds.

## The specific holes the guard covers today

Worth listing, because the guard is the general fix and these are individually fixable:

- **`dist._path`, line 111.** Private, and only on `PathDistribution`. Worth fixing at the source: `dist.locate_file("")` is the public equivalent that every `Distribution` implements.
- **`importlib.metadata.distributions()`, line 108.** Enumerates every distribution on `sys.path` and parses their metadata.
- **`from packaging.requirements import Requirement`, lines 143-144.** Resolves against an application-supplied `packaging` whenever the application has one, because the site is appended last on line 299 so that the application's versions win. An old or broken `packaging` raises `ImportError` here.
- **`import subprocess`, line 205**, and `import logging` if #96 lands, for the same reason: neither is in a bare interpreter's `sys.modules`, so both are shadowable from `PYTHONPATH`.

None of these is exotic enough to leave uncovered, and the guard costs four lines.

ocelotl:
Fixes #98.

`import_distro()` was called with nothing around it, so an exception from any step that has no `try` block of its own escaped into `site.execsitecustomize()`. The process then ended up in the one state every guard in the file exists to prevent, and #93 had already established that this state is worth preventing. #94 closed one door into it; this closes the frame.

## The failure

Reproduced with the packaged script, unmodified, on `python:3.13-slim`. A `.pth` file in `site-packages` replaces `importlib.metadata.distributions` with one returning a `Distribution` that has `metadata["Name"] == "opentelemetry-sdk"` and no private `_path`, which is what line 111 read. The `.pth` is only a way to get in front of `sitecustomize.py`, since `site` processes `site-packages/*.pth` before `execsitecustomize()`.

| | before | after |
|---|---|---|
| site in `PYTHONPATH` | still present | removed |
| `PYTHON_AUTO_INSTRUMENTATION_AGENT_PATH_PREFIX` | still set | cleared |
| `[opentelemetry-python-autoinstrumentation]` lines | 0 | 1 |
| child process repeats the failure | yes | no |
| what the operator sees | `Error in sitecustomize; set PYTHONVERBOSE for traceback:` | the guard message, naming the cause |
| application exit code | 0 | 0 |

Three things made it worth fixing above the other findings in this file:

1. **The report does not name this agent.** CPython prints one line, without the `[opentelemetry-python-autoinstrumentation]` prefix and without a traceback unless `PYTHONVERBOSE` is set. This repository's own container assertions grep for that prefix, so they would not catch it either.
2. **Every child process retries.** `_self_deactivate` exists precisely so children do not, and it never ran. On gunicorn or celery that multiplies by the worker count.
3. **`sys.path` was left inconsistent.** The failure lands between the removal on line 291 and the re-append on line 299, so the parent had neither instrumentation nor the bundle importable. #93's failure at least left the bundle on `sys.path`.

The application was never harmed and always exited 0, which is what let this go unnoticed.

## Commits

1. `fix(python)`: the outer guard. Its own reporting is guarded too, because the step that failed can be the step that reports. The new unit test fails without it, with the `RuntimeError` propagating out of `exec_module`.
2. `fix(python)`: replace the private `dist._path` with `locate_file("")`, removing the most reachable trigger. `locate_file` is abstract on `Distribution`, so every implementation has it, while `_path` exists only on `PathDistribution`. The message gains the version: `opentelemetry-sdk 1.20.0 (/app/site-packages)`.
3. `docs(python)`: the safety-guard list stopped at the four anticipated checks, and never said what deactivating actually does.

The two fixes are complementary, and the second commit's test shows why both are wanted. Without the `locate_file` change, the guard from commit 1 catches the `AttributeError` and deactivates with "unexpected error" rather than correctly reporting the double instrumentation. The guard makes the failure graceful; the `locate_file` change makes it correct.

## Verified

- The unit suite, 32 tests.
- `TestSitecustomizePythonVersionCompatibility`, all four interpreters.
- `python2.7 -m py_compile` on the script, so the parse contract on line 11 holds.
- The reproduction above, before and after, plus a third run proving the `locate_file` path reports the offending distribution correctly rather than merely failing gracefully.

## Notes for review

**The inner guard is deliberate and untested.** Reaching it needs `_self_deactivate` or the diagnostics channel itself to fail, which the unit harness cannot arrange without contortions. I would rather have four lines of unreachable insurance in a file that runs in every Python process on the host than a handler that can raise from inside an exception handler.

**Independent of #96 and #97.** Different lines, no dependency, and either order merges. The only overlap is that both PRs insert into `test_sitecustomize.py`, at different anchors, so a conflict there would resolve by keeping both.

**One interaction worth knowing about.** #97 makes the diagnostics import `logging` lazily, and the two module-level `_log_debug` calls near the top of the file sit outside this guard. If both land, moving those two calls inside `import_distro()` would put them under it, which is a one-line follow-up rather than something to settle here.

**This does not fix the other unguarded call sites, by design.** `importlib.metadata.distributions()` on line 108, `from packaging.requirements import Requirement` on lines 143-144 which resolves against an application-supplied `packaging`, and `import subprocess` on line 205 which is shadowable from `PYTHONPATH`. The guard covers all of them, and each is a separate question about whether it should also be handled at the source.

### DIFF
--- a/packaging/common/python/sitecustomize.py
+++ b/packaging/common/python/sitecustomize.py
@@ -382,4 +382,21 @@ def import_distro():
         _print_cannot_auto_instrument_message("dependency conflicts: {}".format(version_conflicts))
 
 
-import_distro()
+# Every guard above exists to leave the process in a known state, and an
+# exception escaping import_distro() is the one path that leaves it in the
+# state they prevent: the site still on PYTHONPATH and the injector prefix
+# still set, so every child process that execs an interpreter retries the same
+# failure, while site.execsitecustomize() reports a single line that does not
+# name this agent and carries no traceback unless PYTHONVERBOSE is set.
+# The reporting is itself guarded, because the step that failed can be the one
+# that reports.
+try:
+    import_distro()
+except Exception as unexpected_error:
+    try:
+        _self_deactivate(dirname(__file__))
+        _print_cannot_auto_instrument_message(
+            "unexpected error while deciding whether to auto-instrument: {}: {}".format(
+                type(unexpected_error).__name__, unexpected_error))
+    except Exception:
+        pass
--- a/packaging/common/python/sitecustomize.py
+++ b/packaging/common/python/sitecustomize.py
@@ -149,7 +149,11 @@ def _check_for_double_instrumentation(current_site):
     for dist in importlib.metadata.distributions():
         name = dist.metadata["Name"]
         if name is not None and name in double_instrumentation_check_packages:
-            offending_packages.append(str(dist._path))
+            # The operator reading the deactivation message has to find and
+            # remove this package, so name it with its version and its install
+            # directory. locate_file("") is abstract on Distribution, so that
+            # directory resolves for any finder, not just path-based installs.
+            offending_packages.append("{} {} ({})".format(name, dist.version, dist.locate_file("")))
     if offending_packages:
         _self_deactivate(current_site)
         _print_cannot_auto_instrument_message(
--- a/packaging/common/python/README.md
+++ b/packaging/common/python/README.md
@@ -107,6 +107,12 @@ configuration file that `otel-config-check` accepts and the SDK's file configura
 rejects: the validator checks the YAML and `file_format`, while the SDK validates the whole
 document against its schema.
 
+Any unexpected error in the checks themselves self-deactivates as well, and reports the
+exception type and message. Deactivating means the site directory is removed from
+`sys.path` and from `PYTHONPATH`, and the injector's agent path prefix is cleared, so child
+processes do not retry a failed activation. The application is never prevented from
+starting by any of this.
+
 ## Supported libraries
 
 Instrumentation plug-ins are included for:
--- a/packaging/common/python/test_sitecustomize.py
+++ b/packaging/common/python/test_sitecustomize.py
@@ -209,6 +209,7 @@ class ImportDistroTests(unittest.TestCase):
         installed_distributions=None,
         sys_path_entry=None,
         initialize_side_effect=None,
+        distributions_side_effect=None,
     ):
         """Execute sitecustomize.py end to end.
 
@@ -220,6 +221,10 @@ class ImportDistroTests(unittest.TestCase):
 
         initialize_side_effect makes the mocked initialize() raise, which is
         how the SDK reports a failure it cannot recover from.
+
+        distributions_side_effect makes importlib.metadata.distributions raise,
+        which stands in for any failure inside import_distro() that has no
+        try block of its own.
 
         Returns (stderr output, auto_instrumentation mock, environment
         observed right after the run).
@@ -248,7 +253,8 @@ class ImportDistroTests(unittest.TestCase):
         with patch.dict(os.environ, env, clear=True), \
                 patch.object(sys, "path", list(sys.path) + [sys_path_entry or self.site_dir]), \
                 patch("os.path.dirname", side_effect=fake_dirname), \
-                patch("importlib.metadata.distributions", return_value=installed_distributions or []), \
+                patch("importlib.metadata.distributions", return_value=installed_distributions or [],
+                      side_effect=distributions_side_effect), \
                 patch("importlib.metadata.distribution", return_value=distribution), \
                 patch.dict(sys.modules, {
                     "opentelemetry": MagicMock(),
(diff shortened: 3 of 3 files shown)


packaging/common/python/sitecustomize.py
```diff
@@ -108,7 +108,13 @@ def _check_for_double_instrumentation(current_site):
     for dist in importlib.metadata.distributions():
         name = dist.metadata["Name"]
         if name is not None and name in double_instrumentation_check_packages:
-            offending_packages.append(str(dist._path))
+            # dist._path is the .dist-info directory, but it is private and
```
mmanciop: I am not sure these comments are valuable

mmanciop (approved):

packaging/common/python/sitecustomize.py
ocelotl: Right, fixed :v:

mmanciop (approved):

